Issue #22, opened 2026-06-14
Email verification is not enforced for protected routes (advisory only)


Oreginal:
## Summary
After signup, an account is created and the session cookie is set even though the email is unverified. Nothing blocks an unverified user from reaching `/account` or `/checkout` — `proxy.ts` and `AccountContent` only check session presence, and the account page shows a soft "Please verify your email" reminder.

This was an intentional Phase-1 decision (verification is advisory), but it's recorded here so it's a conscious choice rather than an oversight.

## Why it matters
If business logic later assumes "reached /checkout ⇒ verified email" (e.g. for orders/payments), that invariant is currently unenforced.

## Options
1. **Keep advisory** (close as wontfix) — fine if verification is never a hard gate.
2. **Enforce at checkout only** — block `/checkout` until `user.emailVerified`, allow `/account`.
3. **Enforce broadly** — gate all protected routes on `emailVerified`.

Note: `emailVerified` is not available to the cookie-based `proxy.ts` (edge, no Firebase session), so enforcement must happen client-side in the protected page content (like the existing stale-session guards) or via a verified Firebase session cookie (Admin SDK) if stronger guarantees are needed.

## Locations
- `src/proxy.ts`
- `src/app/account/AccountContent.tsx`
- `src/app/checkout/CheckoutContent.tsx`

## Acceptance criteria
- [ ] Decision made (advisory vs enforced) and documented
- [ ] If enforced: unverified users are redirected to `/verify-email` from the gated route(s)

_Surfaced during the auth + cart code review (deferred decision)._

Oreginal:
**Decision: enforce at `/checkout` only** (option 2). `/account` remains advisory (soft reminder retained).

Rationale: the invariant that actually matters is "reached checkout ⇒ verified email" (orders/payments). Gating the account dashboard too would lock unverified users out of their own account for little benefit.

Implemented in #49:
- `/checkout` gated on `emailVerified` client-side in `CheckoutContent` (the edge proxy can't read Firebase `emailVerified`), redirecting unverified users to `/verify-email?from=<dest>` with the full destination preserved.
- `/verify-email` auto-detects verification (poll + tab-focus via new `AuthContext.reloadUser()`) and returns the user to `<dest>`.

Acceptance criteria:
- [x] Decision made (enforce at checkout) and documented — here, in `CLAUDE.md`, and in `proxy.ts`.
[...]

Issue #3, opened 2026-05-30
Product Detail Pages


Oreginal:
### Create dynamic product pages driven by Shopify Date
Acceptance Criteria:

- [x] Product images gallery
- [x] Product information
- [x]  Variants
- [x] Add to cart 
- [x] Related/Recommended products
- [x] Customer product review

Oreginal:
## Progress — Store + product detail pages (placeholder data)

Implemented on branch `oreutlwile/feat/3/product-detail-pages` (commit 879195b). Built against placeholder data in `src/lib/products.ts`, structured as the single **Shopify Storefront API swap point**.

**New routes**
- `/store` — catalog listing: all products grid + curated bundles section.
- `/store/[slug]` — dynamic product detail page (SSG via `generateStaticParams` + per-product `generateMetadata`, `notFound()` for bad slugs).

**Acceptance criteria**
- [x] Product images gallery — main image + thumbnail strip (click to swap)
- [x] Product information — name, category, rating, price (+ original), description, tags
- [x] Variants — selectable chips that drive the displayed price
- [x] Add to cart — variant-aware cart lines (`product:N#<variant>`); wishlist toggle too
- [x] Related/Recommended products — same-category-first grid of product cards
- [x] Customer product review — review list with average summary, **filterable by star rating**; rating stars in the product header are clickable and smooth-scroll to reviews

**Also**
- Extracted reusable `ProductCard`/`BundleCard`; homepage Collection + Bundles now share the same data/cards. Product cards link to their detail page.
- Navbar "Shop" → "Store" (`/store`); "View Full Collection" and the Shop CTAs route to `/store`; homepage anchors made route-aware.

**Why this stays open:** all criteria are met with **placeholder** data only. Closing is gated on the real Shopify Storefront API integration (replacing `src/lib/products.ts` data + types with live products/variants/inventory, and persisting real reviews) under the *Shopify Ecommerce Integration* milestone.

Issue #2, opened 2026-05-30
Integrate Shopify Storefront


Oreginal:
### Connect the frontend to Shopify products, collections, pricing, inventory, and checkout flow.
Acceptance Criteria:

- [ ] Products load from Shopify 
- [ ] Collections load dynamically
- [ ] Product detail pages populated from Shopify
- [ ] Checkout redirects correctly

PR #23, opened 2026-06-15
Auth pages, cart system, and cart tech-debt cleanup (#19, #20, #21)

### LINKED ISSUE #1: Auth Pages

### LINKED ISSUE #19: Extract shared SideDrawer primitive (CartSidebar duplicates WishlistSidebar)
## Summary
`src/components/CartSidebar.tsx` is a near-verbatim copy of `src/components/WishlistSidebar.tsx`. The backdrop, slide-in panel wrapper, header (animated icon + title + count badge + close button), empty-state block, per-item card chrome, and footer are structurally identical — only the line items, footer (subtotal + checkout), and labels differ.

## Why it matters
Any change to the drawer shell (z-index, blur value, animation timing, close-button a11y, responsive width, focus trap, body scroll-lock) must be made and tested in two places, and they will drift. ~250 lines are duplicated.

## Suggested fix
Extract a `<SideDrawer side="right">` primitive that owns the backdrop + panel + header/footer slots, with the body passed as `children`. Refactor both `CartSidebar` and `WishlistSidebar` to use it.

## Locations
- `src/components/CartSidebar.tsx`
- `src/components/WishlistSidebar.tsx`

## Acceptance criteria
- [x] Shared drawer primitive created
- [x] Both sidebars use it with no visual/behaviour regression
- [x] Duplicated shell code removed

_Surfaced during the auth + cart code review (deferred as tech debt)._

### LINKED ISSUE #20: Bundle cart IDs use a fragile magic-number offset (collision risk)
## Summary
Bundles are given cart IDs `1001`–`1004` (in `src/components/sections/Bundles.tsx`) chosen to avoid colliding with product IDs `1`–`5`. The non-collision is enforced only by a code comment and human discipline, not by the type system or data model.

## Why it matters
`CartContext.addItem` keys on `id`. When the catalog grows (products seeded from a DB with auto-increment IDs that cross 1000, or more bundles added), a bundle `cartId` can silently equal a product `id`. Adding that bundle would then increment the quantity of an unrelated product (and vice versa), corrupting the cart with no error.

## Suggested fix
Use namespaced keys so collisions are structurally impossible, e.g. `product:3` vs `bundle:glow`. This means changing `CartProduct.id` (and the wishlist's `id`) from `number` to a namespaced `string`, and updating `isInCart`/`removeItem`/`updateQuantity` callers.

## Locations
- `src/components/sections/Bundles.tsx` (cartId 1001–1004)
- `src/context/CartContext.tsx` (id-keyed logic)
- `src/components/sections/FeaturedProducts.tsx` (product ids)

## Acceptance criteria
- [x] Cart/wishlist item identity is namespaced and collision-proof
- [x] No magic-number offsets remain

_Surfaced during the auth + cart code review (deferred as tech debt)._

### LINKED ISSUE #21: Replace hardcoded color/easing literals with theme tokens
## Summary
Brand colours and the cinematic easing curve are hardcoded as literals throughout the new auth/cart components (and the pre-existing codebase): `#c9977a` (~41×), `rgba(201,151,122,…)` (~38×), `#1e1814`, and the easing array `[0.16, 1, 0.3, 1]` (~9×).

`tailwind.config.ts` already defines `rose.gold = #c9977a`, `cream`, `parchment`, and `box-shadow` tokens; `src/lib/motion.ts` already exports `ease.cinematic = [0.16, 1, 0.3, 1]`.

## Why it matters
A brand-colour or motion refresh requires a find-and-replace across ~80 literal occurrences instead of editing one token; any missed instance leaves an off-brand value. Note `#1e1814` (primary ink) isn't a token at all — there's no single place to define it.

## Suggested fix
- Add an `ink`/`foreground` token for `#1e1814` to the Tailwind theme.
- Replace literal colours with token classes / CSS vars.
- Import `ease` from `src/lib/motion.ts` instead of inlining the easing array.

## Note
This is pre-existing convention across the whole codebase (Navbar, WishlistSidebar, FeaturedProducts all inline literals), so it's a broader cleanup, not isolated to the auth/cart work.

## Acceptance criteria
- [ ] Brand ink colour tokenised
- [ ] Colour/easing literals replaced with tokens in auth/cart components

_Surfaced during the auth + cart code review (deferred as tech debt)._

Oreginal:
## Summary

Brings the Firebase auth flow + cart system to `master`, plus the first round of post-review tech-debt cleanup.

### Auth + cart (commits `57f962d`, `64703da`)
- Firebase auth (Email/Password + Google + Apple) under an `(auth)` route group: login, signup (password-strength meter), forgot-password, verify-email.
- `AuthContext` with cookie-flag session signalling; `proxy.ts` gates `/account` + `/checkout` and bounces signed-in users off auth pages.
- Cart system (`CartContext` + `CartSidebar` + `CartNotification`) with localStorage persistence, quantity handling, and an auth-gated checkout. Wired into FeaturedProducts, Bundles, and the navbar bag badge.
- Protected `/account` + `/checkout` stubs.

### Tech-debt cleanup (commit `17d375c`)
- **#20 — Namespaced ids:** cart/wishlist item `id` is now a `string` (`product:N`, `bundle:<slug>`), removing the fragile `1001–1004` magic offset. `CartContext` drops legacy numeric-id entries from `hb-cart` localStorage.
- **#19 — Shared `<SideDrawer>`:** new primitive owns the backdrop/panel/header/footer shell; `CartSidebar` + `WishlistSidebar` compose it instead of duplicating ~250 lines. No visual/behaviour change.
- **#21 — Tokenize (auth/cart scope):** added `ink` (`#1e1814`) Tailwind token; replaced brand-colour class literals (`text-rose-gold`/`text-ink`) and inline cinematic easing (`ease.cinematic`) across the drawer + auth/cart components.

Closes #19, #20, #21.

## Firebase status
Real credentials populated; all three sign-in methods enabled in the console; Google sign-in manually verified working.

## Verification
- `tsc --noEmit` clean.
- `next build` compiles + typechecks clean.
- Dev-server route probes: public/auth pages 200; `/account` + `/checkout` 307 (proxy gating intact).
- ⚠️ Interactive cart click-through (add/qty/remove, bundle no-collision) not yet browser-tested.

## Known issues (not introduced here)
- `next build` fails prerendering the synthetic `/_global-error` page (`InvariantError: Expected workStore to be initialized`, Next 16.2.6 + Turbopack). Confirmed failing on the clean tree at `64703da` — pre-existing, blocks a full production build, deserves its own fix.

## Deferred
- **#22** (enforce email verification) — decision made: enforce at checkout only. Not implemented in this PR.

## Release Notes

* **New Features**
  * Complete authentication flow: sign up, sign in, password reset, email verification (including social sign-in buttons).
  * Protected account and checkout pages with redirect-safe sign-in behavior.
  * Storefront with product detail pages, related recommendations, and curated bundles.
  * Interactive cart sidebar, “added to bag” notifications, and wishlist items with “move to bag” actions.
  * Account dashboard plus custom global error and 404 pages.

* **Bug Fixes**
  * Improved page metadata and routing behavior across key auth and product pages.

* **Documentation**
  * Added environment configuration template and project guide.


Oreginal:
## Code Review — PR #23: Auth, Cart & Cart Tech-Debt

---

### 🔴 High

**1. After signInWithRedirect (popup-blocked / iOS), user is never navigated to their destination**
`src/app/(auth)/login/page.tsx` — `handleSocial` / `src/context/AuthContext.tsx` — `socialSignIn`

When a popup is blocked, `socialSignIn` falls back to `signInWithRedirect(auth, provider)`. The browser immediately navigates away to Google — `handleSocial`'s `router.push(destination)` fires but is a no-op on an unloading page. When the user returns, `getRedirectResult` sets the cookie and `onAuthStateChanged` sets `user`, but `LoginForm` has no `useEffect` watching `user` — it never reacts to the sign-in completing. The user sits on `/login` while fully authenticated. Affects every iOS user and any browser with strict popup settings.

**Fix:** Add to `LoginForm`:
```typescript
const { user } = useAuth();
useEffect(() => {
  if (user) router.push(destination);
}, [user, destination, router]);
```
And persist `destination` to `sessionStorage` before calling `signInWithRedirect` so it survives the page reload.

---

**2. `?from=` redirect param is lost through the redirect OAuth flow**
`src/app/(auth)/login/page.tsx` + `src/proxy.ts`

The proxy correctly sets `?from=/checkout` when redirecting to `/login`. For redirect sign-in, the browser leaves and Firebase returns the user to `/login` without the `?from=` param — Firebase's redirect handler only preserves the base URL. The user ends up at `/account` instead of `/checkout`.

**Fix:** Before calling `signInWithRedirect`, store the destination in `sessionStorage`. Restore and navigate to it inside the `useEffect` from fix #1.

---

**3. localStorage cart restore missing `Array.isArray` guard — non-array JSON silently destroys the cart**
`src/context/CartContext.tsx` (~line 55)

If a previous app version stored `{ items: [...] }` under the same key, `JSON.parse` succeeds and returns a plain object. `.filter` throws `TypeError` on objects. The `catch {}` swallows it, `setLoaded(true)` runs, and the persist effect immediately writes `[]` back — permanently destroying the cart.

**Fix:**
```typescript
const parsed = JSON.parse(raw);
if (Array.isArray(parsed)) {
  const restored = parsed.filter((p) => typeof p.id === "string");
  setItems(restored);
}
```

---

### 🟠 Medium

**4. CartNotification motion key changes from number to string on dismiss, aborting exit animation**
`src/components/CartNotification.tsx` (~line 28)

```tsx
<motion.div key={lastAdded?.key ?? "cart-notification"} ...>
```

When the timer fires, `clearLastAdded()` sets `lastAdded` to null — `lastAdded?.key` becomes `undefined` and the key flips to the string `"cart-notification"`. Framer Motion sees a key change, unmounts and remounts the element mid-exit, hard-cutting the animation.

**Fix:** Capture the key when `shown` is set and hold it stable — drive the motion key off a local state variable rather than `lastAdded?.key`.

---

**5. Price rounding inconsistency — line items use `.toFixed(0)`, total uses `.toFixed(2)`**
`src/components/CartSidebar.tsx` + `src/app/checkout/CheckoutContent.tsx`

A product at R149.99 shows as `R150` per line but the total reads `R149.99`. Two such items imply R300 from the lines but R299.98 as the total. Visible to every customer with non-whole-rand prices.

**Fix:** Use `.toFixed(2)` consistently on all price displays.

---

**6. `AUTH_COOKIE` constant defined separately in two files — rename in either silently breaks auth for all users**
`src/proxy.ts:3` + `src/context/AuthContext.tsx`

Both define `const AUTH_COOKIE = "hb-auth-token"` independently. If a developer renames it in one file without the other, every authenticated user is bounced into a redirect loop with no TypeScript error surfacing it.

**Fix:** Extract to `src/lib/constants.ts` and import in both files.

---

**7. Open redirect via backslash — `resolveDestination` blocks `//evil.com` but not `\/evil.com`**
`src/app/(auth)/login/page.tsx` — `resolveDestination`

`\/evil.com` starts with `/` and not `//`, so it passes the current guard. Chromium treats this as an external URL. An attacker crafting `/login?from=\/evil.com` redirects users off-site after login.

**Fix:** Add `!from.startsWith("/\\")` to the guard, or validate against an allowlist of known safe paths.

---

**8. `$` hardcoded as currency symbol across cart, checkout, and notification — wrong for a ZAR store**
`src/components/CartSidebar.tsx`, `CartNotification.tsx`, `src/app/checkout/CheckoutContent.tsx`

Every price shows `$299.00` instead of `R299.00`. Wrong today and permanently wrong across all price touchpoints when PR #33 lands Shopify ZAR amounts.

**Fix:** Shared formatter:
```typescript
export const formatPrice = (amount: number) =>
  new Intl.NumberFormat("en-ZA", { style: "currency", currency: "ZAR", minimumFractionDigits: 0 }).format(amount);
```

---

**9. `setLastAdded` always passes `quantity: 1`, even when re-adding an existing cart item**
`src/context/CartContext.tsx` (~line 100)

`setItems` correctly increments to 3 on a re-add, but the notification toast always shows quantity 1.

**Fix:**
```typescript
const newQty = existing ? existing.quantity + 1 : 1;
setLastAdded({ product: { ...product, quantity: newQty }, key: Date.now() });
```

---

### 🔵 Low

**10. Auth `page.tsx` files use `"use client"` directly — CLAUDE.md violation, blocks `metadata` exports**
`src/app/(auth)/login/page.tsx`, `signup/page.tsx`, `forgot-password/page.tsx`, `verify-email/page.tsx`

CLAUDE.md: *"a top-level route is a server `page.tsx` that renders a `"use client"` `*Content.tsx` for interactivity."* A `"use client"` `page.tsx` cannot export `metadata` in the App Router — adding per-page titles or Open Graph tags will silently fail or throw a build error.

---

**11. Inlined easing `[0.16, 1, 0.3, 1]` in CartNotification instead of `ease.cinematic`**
`src/components/CartNotification.tsx` (~lines 40, 58)

CLAUDE.md: *"import variants + easing from `src/lib/motion.ts` … rather than inlining `[0.16, 1, 0.3, 1]`."* `CartSidebar.tsx` in the same PR correctly uses `ease.cinematic` — `CartNotification` should match.

---

### Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 High | `login/page.tsx` + `AuthContext.tsx` | Redirect OAuth: user stranded on login, no post-auth navigation |
| 2 | 🔴 High | `login/page.tsx` + `proxy.ts` | `?from=` param lost through redirect flow — wrong destination post-login |
| 3 | 🔴 High | `CartContext.tsx` | Missing `Array.isArray` guard — corrupt localStorage silently destroys cart |
| 4 | 🟠 Medium | `CartNotification.tsx` | Motion key flips on dismiss — exit animation hard-cut |
| 5 | 🟠 Medium | `CartSidebar.tsx` + `CheckoutContent.tsx` | `.toFixed(0)` vs `.toFixed(2)` — line items don't visually sum to total |
| 6 | 🟠 Medium | `proxy.ts` + `AuthContext.tsx` | `AUTH_COOKIE` duplicated — silent breakage if one file is renamed |
| 7 | 🟠 Medium | `login/page.tsx` | Backslash open redirect bypasses `//` guard |
| 8 | 🟠 Medium | `CartSidebar`, `CartNotification`, `CheckoutContent` | `$` hardcoded for a ZAR store — wrong on every price |
| 9 | 🟠 Medium | `CartContext.tsx` | `setLastAdded` always `quantity: 1` — wrong toast on re-adds |
| 10 | 🔵 Low | `(auth)/*/page.tsx` | CLAUDE.md: `"use client"` in `page.tsx` blocks `metadata` exports |
| 11 | 🔵 Low | `CartNotification.tsx` | CLAUDE.md: inline `[0.16,1,0.3,1]` — should use `ease.cinematic` |

Three blockers before merge: **#1** (mobile OAuth broken), **#2** (wrong post-login destination), **#3** (cart data loss risk). The rest can be fixed in the same branch. 🚀

PR #33, opened 2026-06-20
Katleho/feat/02/integrate shopify storefront

### LINKED ISSUE #2: Integrate Shopify Storefront

KatlehoMadaba:
closes #2 

## Summary by CodeRabbit

* **New Features**
  * Featured products now load from Shopify at render time and populate the product grid.
  * Product cards updated to use Shopify pricing (conditional “Sale” label, original/strikethrough pricing) and improved tag display; rating UI removed.
* **Bug Fixes**
  * Suppressed newsletter email hydration mismatch warnings.
  * Wishlist sidebar now formats item prices consistently and no longer shows item category text.
* **Documentation**
  * Added a comprehensive Storefront API reference (authentication, directives, rate limits, errors).
* **Chores**
  * Allowed images from the Shopify CDN; updated build/dev scripts and dependencies.


Oreginal (commented):
Hey Kat! Nice work on this pulling the featured products from the live Storefront API and rendering real data is a great step. I tested the endpoint and confirmed the tokenless query works and returns the catalog in ZAR, so the core integration is solid. A couple of production-hardening things before we merge:

### 1. 🔴 Add caching/revalidation to the fetch (most important)

Tokenless Storefront access is rate-limited much more tightly than tokenized access. Right now the `fetch` is uncached, so it re-runs on **every** homepage render — under real traffic Shopify will start returning 429s and the section silently goes empty. Caching it once an hour fixes this and makes the page faster:

```ts
const response = await fetch(API_URL, {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Accept': 'application/json',
  },
  body: JSON.stringify({ query: PRODUCTS_QUERY }),
  next: { revalidate: 3600 }, // cache 1h; products refresh hourly
});
```

### 2. 🔴 Wrap the fetch so a network error can't crash the homepage

We handle HTTP errors and GraphQL errors (👍), but if `fetch` itself throws (network blip, bad URL), it bubbles up through the `async Home()` server component and 500s the whole page — not just this section. Since there's no `error.tsx` boundary yet, let's catch it and degrade to an empty list:

```ts
export async function getFeaturedProducts() {
  try {
    const response = await fetch(API_URL, { /* ...as above... */ });

    if (!response.ok) {
      console.error('Shopify fetch error:', response.status, await response.text());
      return [];
    }

    const json = await response.json();
    if (json.errors || !json.data?.products) {
      console.error('Shopify query errors:', json.errors);
      return [];
    }

    return json.data.products.edges.map(/* ...unchanged... */);
  } catch (err) {
    console.error('Shopify fetch failed:', err);
    return [];
  }
}
```

(Note the small extra `!json.data?.products` guard — protects the `.edges.map` from a null `data`.)

### 3. 🟡 Minor cleanups (non-blocking)

- **`src/lib/shopify.ts`** — this new file isn't imported anywhere (the query uses raw `fetch`). Let's either wire it in or delete it so it doesn't confuse the next person.
- **Imageless products** — `image: … ?? ''` will throw in `next/image` ("empty string passed to src") if a product ever has no image. Every current product has one, so this is just future-proofing — a placeholder path as the fallback would cover it.
- **BOM characters** — `FeaturedProducts.tsx` and `shopify-queries.ts` got saved with a UTF-8 BOM at the top. It works, but re-saving as "UTF-8 without BOM" keeps diffs clean.
- **`({ node }: any)`** — would be nice to type the node shape so a schema change doesn't fail silently. Totally optional.

Really nice touches on the `Intl.NumberFormat` price formatting, the `remotePatterns` config, and the `suppressHydrationWarning` fix. Just the two top items needed before merge — the rest is polish. 🙌

Oreginal:
**Review: Shopify Storefront Integration**

Great progress getting the Storefront API client set up and the product data flowing into `FeaturedProducts`. A few things to resolve before this is ready to merge:

---

**🔴 Build blocker — `category` excess-property error**

`WishlistProduct.category` was commented out of the interface, but `FeaturedProducts.tsx` still passes `category: product.category` inside the `toggleItem` call, and `WishlistSidebar.tsx` still renders `{item.category}`. TypeScript strict mode will throw an excess-property error and fail the build.

Fix: Remove `category` from the `toggleItem` call and from the `WishlistSidebar` render. We'll wire the full wishlist↔Shopify integration up in a separate pass once all PRs are merged.

---

**🟡 Incorrect "Sale" badge condition**
`src/lib/shopify.ts`

```ts
// current — wrong
originalPrice: compareAmount > 0 ? compareAmount : null,
```

Shopify can return a non-zero `compareAtPriceRange` value even when the compare-at price equals the selling price, which would show a "Sale" badge and strikethrough on a product that isn't discounted.

```ts
// fix
const price = parseFloat(node.priceRange.minVariantPrice.amount);
originalPrice: compareAmount > price ? compareAmount : null,
```

---

**🔵 Fallback image is a specific product photo**
`src/lib/shopify.ts:85`

```ts
image: node.images.edges[0]?.node.url ?? "/images/item-1.jpeg",
```

`/images/item-1.jpeg` is the Glow Collagen Blend product image. A Shopify product with no images (draft, archived) would render with the wrong product's photo. Swap for a neutral placeholder.

---

**ℹ️ Notes / no action needed**

- The `id` type mismatch between `ShopifyProduct` (`number`) and `WishlistProduct` (`string`) is a known gap — being addressed in the wishlist integration PR.
- ISR revalidation via `fetchApiOptions: { next: { revalidate: 3600 } } as RequestInit` may not thread through the Storefront client's internal fetch — worth revisiting once we confirm caching behaviour in staging.
- `suppressHydrationWarning` on the Newsletter input: the root cause (likely a password manager injecting attributes) should be investigated rather than suppressed, but not a blocker for this PR.

---

Fix the two items above and this is good to go. 🚀

PR #41, opened 2026-07-07
feat: premium full-screen loader for home hero video

### LINKED ISSUE #39: Hey Beautiful logo loader

Oreginal:
## Summary
- Adds a full-screen branded loader (`PageLoader`) that masks the home page until the hero background video decodes its first frame, then reveals with a slide-up unveil
- New `HomeHero` orchestrator wires loader visibility to video readiness (once per session via `sessionStorage`, hydration-safe), with a 5s safety fallback and full `prefers-reduced-motion` support
- `Hero` gains optional `onVideoReady` / `startEntrance` props (backward compatible) and preloads the hero video at high priority via React's `preload()` so the download starts with the initial HTML
- Cleaned up shared `fadeUp`/`fadeIn`/`heroEntrance` motion variants to clear `filter` to `none` after the blur-in entrance completes (GPU/compositing cleanup, no visual change)

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npm run build` succeeds
- [x] First visit: loader shows, scroll locked, unveils once video is ready, hero entrance plays during reveal
- [x] Repeat visit / client-side nav back to home: no loader flash, no hydration mismatch
- [x] Video load failure (renamed source) and slow network: 5s fallback dismisses the loader
- [x] `prefers-reduced-motion`: breathing/glow/slide disabled, plain fade only
- [x] Verified `<link rel="preload" as="video">` present in SSR'd head
- [x] Verified no lingering `filter`/`will-change` on the hero subtree post-reveal

Closes #39

https://claude.ai/code/session_01WM35hgfxx3eoAE9495XKTq


src/components/ui/PageLoader.tsx
```diff
@@ -0,0 +1,150 @@
...
+    // Make the obscured page content inert (not focusable / not in the a11y
+    // tree) and flag it busy for assistive tech while the loader covers it.
+    root?.setAttribute("inert", "");
+    root?.setAttribute("aria-busy", "true");
+
+    return () => {
+      document.body.style.overflow = previousOverflow;
+      document.body.style.paddingRight = previousPaddingRight;
+      root?.removeAttribute("inert");
+      root?.setAttribute("aria-busy", "false");
```
KatlehoMadaba: Nit: cleanup restores `inert` with `removeAttribute` but resets busy with `setAttribute("aria-busy", "false")`, leaving a permanent `aria-busy="false"` on the root. Prefer `root?.removeAttribute("aria-busy")` for symmetry and a clean DOM.

src/components/sections/HomeHero.tsx
```diff
@@ -0,0 +1,89 @@
...
+    // return early without arming a new one.
+    if (skipped || videoReady) return;
+    const seen = hasShownLoader();
+    shownThisSpaSession = true;
+    if (seen) {
+      // Loader is already display:none via the injected style — drop it silently.
+      setSkipped(true);
+      return;
+    }
+    markLoaderShown();
```
KatlehoMadaba, line 62: First-visit scroll-lock relies on PageLoader's effect running before this `markLoaderShown()` (children mount-effects fire before parents). Correct today and documented, but it's an implicit contract — a future wrapper/`Suspense` boundary could silently drop the lock + inert on first visit with no error. Consider driving both the lock and the flag-write from this orchestrator so ordering isn't load-bearing. Not a bug, just fragile.

src/components/sections/HomeHero.tsx
```diff
@@ -0,0 +1,89 @@
...
+const FALLBACK_MS = 5000;
+
+// Survives client-side route changes; resets on full reload, where
+// sessionStorage (checked pre-paint by the inline script) takes over.
+let shownThisSpaSession = false;
+
+// Runs during HTML parse, before first paint — hides the SSR'd loader on
+// repeat visits so it never flashes. Injects a <style> rather than touching
+// React-managed attributes (a class on <html> triggers a hydration mismatch).
+const skipScript = `try{if(sessionStorage.getItem("${LOADER_SESSION_KEY}")){var s=document.createElement("style");s.textContent="[data-page-loader]{display:none!important}";document.head.appendChild(s)}}catch(e){}`;
```
KatlehoMadaba, line 28: Minor: this appends a `<style>` to `<head>` on every hard load where the flag is set and never removes it. Harmless (you want the loader hidden), just noting the small accumulation. Also — if a CSP with `script-src`/`style-src` is ever added, both this inline script and the injected style would need a nonce.

src/components/sections/HomeHero.tsx
```diff
@@ -0,0 +1,89 @@
...
+import PageLoader from "@/components/ui/PageLoader";
+import {
+  LOADER_SESSION_KEY,
+  hasShownLoader,
+  markLoaderShown,
+} from "@/lib/pageLoader";
+import { ease } from "@/lib/motion";
+
+/** How long to wait for the video before revealing anyway. */
+const FALLBACK_MS = 5000;
```
KatlehoMadaba, line 16: Confirm the feel: a slow network holds the branded loader up to 5s before force-revealing. The `error` listener covers hard failures fast, so this only bites on genuinely slow connections — just checking 5s is the intended patience ceiling.

KatlehoMadaba (commented):
Reviewed the full diff. Overall this is high quality — SSR preload hoisting, no-flash repeat visits, scroll-lock, a11y, and guaranteed dismissal are all handled deliberately and correctly. A few notes below; nothing blocking.

**Also (not line-anchored):** `logo-hearts-white.png` was committed to **both** `public/images/` (served) and repo-root `images/` (not served), byte-identical. The root copy is 100 KB of dead weight — please remove it. Bonus: the source is 2956×2817 but renders at 140–180px, so a pre-downsized asset would shrink the `priority` download.

**Nice work on:** `preload()` in the render body, the `onVideoReadyRef` pattern to avoid effect re-runs, the `loadeddata`+`readyState`+`error` coverage with full cleanup, reduced-motion paths throughout, scrollbar-width compensation on scroll-lock, and the portaled `role=status` announcement outside the inert subtree.

src/components/ui/PageLoader.tsx
Oreginal: Fixed in 5ae9861 — cleanup now uses `root?.removeAttribute("aria-busy")` to match the `inert` removal, so no stale `aria-busy="false"` is left on the root.

src/components/sections/HomeHero.tsx
Oreginal: Good call — refactored in 5ae9861 so the ordering is no longer load-bearing. The scroll-lock + `inert`/`aria-busy` moved into a new `engagePageLock()` helper in `lib/pageLoader.ts`, and `HomeHero` now reads/writes the session flag **and** engages the lock in the same effect. Since one component owns both the read and the write, a future `Suspense`/wrapper boundary reordering effects can no longer drop the lock.

`PageLoader` is now purely presentational. One subtlety I kept: the lock releases on `AnimatePresence`'s `onExitComplete` (after the 0.8s slide-up), not when `videoReady` flips — so the page stays locked while the loader animates away, exactly as the old unmount-timed release did. Release is idempotent, with an unmount-only effect as a safety net.

src/components/sections/HomeHero.tsx
Oreginal: Left as-is for now, on purpose. The accumulation is harmless here — we *want* `[data-page-loader]` hidden on repeat loads, so a lingering `<style>` is a no-op rather than a leak, and it only appears on hard reloads where the flag is set. I didn't want to add teardown complexity for zero behavioural gain.

The CSP point is fair though — noting it explicitly so it isn't forgotten: **if we ever add a CSP with `script-src`/`style-src`, both this inline `<script>` and the injected `<style>` will need a nonce.** Happy to file a follow-up issue to track that. If you'd rather I make the injected style self-removing now for cleanliness, say the word and I'll add it.

src/components/sections/HomeHero.tsx
Oreginal: Confirmed — 5s is the intended patience ceiling. The `error` listener fast-paths genuine failures, so this only bites on a real slow connection where we'd rather reveal the hero than hold the branded loader indefinitely. It's a single documented constant (`FALLBACK_MS`), so easy to retune if it feels long in the wild. Leaving at 5s for now.

Oreginal:
Thanks for the thorough review 🙏 Addressed in 5ae9861:

- **Duplicate asset** — removed the byte-identical root `images/logo-hearts-white.png`; the served copy in `public/images/` is retained.
- **`aria-busy` nit** — cleanup now `removeAttribute`s it (replied inline).
- **Effect-ordering fragility** — refactored so `HomeHero` owns the lock + flag; ordering is no longer load-bearing (replied inline).
- **Injected `<style>` / CSP** and **5s ceiling** — replied inline (kept as-is with reasoning; CSP noted for a follow-up).

One follow-up you flagged I did **not** do: the loader source is 2956×2817 but renders at 140–180px. Pre-downsizing that `priority` asset would shrink the first-load download — I'll open a separate issue for it rather than scope-creep this PR. Shout if you'd prefer it in here.

Oreginal:
Filed follow-ups for the two items scoped out of this PR: #45 (downsize loader logo asset) and #46 (CSP nonce for the inline script/style).

Oreginal:
@KatlehoMadaba requesting re-review — all feedback from your pass is addressed in 5ae9861 (see inline replies above) plus follow-up issues #45 and #46 for the two items scoped out. Ready whenever you are.

PR #43, opened 2026-07-07
Updating Claude.md and env.example

Oreginal:


KatlehoMadaba (commented):
## Review — `main → master` production promotion

Heads up that despite the title ("Updating Claude.md and env.example"), this is a **1054-line production promotion** bundling five unrelated changes: the branching-strategy docs, a full-screen page loader, a new Testimonials section, a shared `ShopifyProductCard` extraction, and a Shopify revalidation webhook.

The code itself is well-crafted — the loader orchestration comments are genuinely good, the `revalidateTag(tag, profile)` two-arg call is correct for Next 16.2 (the profile arg is required in the type defs), and the a11y work in `PageLoader` (portaled `role="status"`, `inert` on the content root, reduced-motion branches) is more careful than most. My concerns are mostly about **what's being shipped to production**, not how it's written.

### Blocking (see inline comments)

1. **Fabricated "Verified Buyer" testimonials** — invented names, cities, purchase dates and verified badges, deploying straight to production. Consumer-protection exposure (SA CPA), not just an ethical one.
2. **Empty-string secret bypasses `/api/revalidate`** — if `SHOPIFY_REVALIDATE_SECRET` is defined-but-empty (exactly what `.env.example` ships), `?secret=` authenticates and the cache purge endpoint is open to anyone.
3. **Placeholder catalog reintroduced as the production fallback** — a misconfigured env var now silently ships a fake storefront instead of failing visibly.

### Significant

4. **Blocking the home page on a 3.7 MB video** — up to a 5s blank brand screen on slow connections. Bad LCP, and LCP feeds SEO ranking. Also `<link rel="preload" as="video">` is not a supported destination in Chromium.
5. **Blank-hero flash on repeat visits** — `useEffect` runs after paint, so returning visitors get a frame with the loader hidden *and* the hero still at its `hidden` variant.
6. **PR hygiene** — the title describes ~15 lines of the diff, and the body is empty. If squash-merged, `master`'s history records a page loader, a testimonials section and a webhook endpoint as "Updating Claude.md and env.example." Per CLAUDE.md, `main → master` should be production-ready only, and a reviewer approving this has no summary of what's being promoted.
7. **The CLAUDE.md being updated is stale in the same PR** — its opening paragraph still says "Shopify isn't wired up yet, so product/bundle data currently lives in `src/lib/products.ts` as placeholders," while this very PR adds a Shopify revalidation webhook and tag-based caching. Worth fixing here.

### Minor

- **`NEXT_PUBLIC_` on the Storefront token is unnecessary.** `getFeaturedProducts` is only called from a server component. Storefront tokens are designed to be public, but the prefix inlines it into the client bundle for no reason.
- **Hex literals over design tokens.** `ShopifyProductCard` uses `#c9977a` / `#1e1814` throughout; `Testimonials` mixes `text-rose-gold` with raw `#c9977a` and `#e8c4ad`. CLAUDE.md says prefer the tokens. Mostly carried over from the deleted code — but the extraction was the moment to clean it up.
- **Duplicated `ZAR` formatter** — worth one export in `src/lib`.
- **`PageLoader` logo is a 2956×2817 PNG (97 KB) rendered at 140–180px**, marked `priority`, so it competes with the hero video for bandwidth during exactly the window you're trying to shorten. Ship a ~360px asset.
- **Test coverage** — none, and nothing in the repo to hang tests on. The revalidate route's auth check is exactly the kind of bug a three-line test catches.

### Recommendation

I'd hold items 1–3 out of production. I'd also suggest splitting this: loader + testimonials as a feature PR, Shopify revalidation as an infra PR — promoting them as one commit titled after the docs change makes both hard to revert independently.

src/components/sections/Testimonials.tsx
```diff
@@ -0,0 +1,334 @@
...
+  span: string;
+  offset: string;
+  entrance: Entrance;
+}
+
+// Self-contained like the sibling sections (Reviews `reviews`, Lifestyle
+// `editorialMoments`). Images are candid lifestyle photography in
+// public/images/testimonials — swap paths here to change them.
+const testimonials: Testimonial[] = [
+  {
```
KatlehoMadaba, line 51: **Blocking — fabricated testimonials labelled "Verified Buyer".**

These four customers are invented: full names, occupations, cities, "2 weeks ago", a `BadgeCheck` **"Verified Buyer"** badge, and "Purchased — Glow Collagen Blend".

This PR deploys straight to production via Netlify. Presenting fabricated reviews as verified purchases is a consumer-protection problem in most jurisdictions (South Africa's CPA included), not only an ethical one. It also compounds the placeholder-catalog issue in `shopify.ts` — the products these customers claim to have purchased are the *placeholder* entries, so they reference SKUs that don't exist.

If the section must ship before real reviews exist, drop the `Verified Buyer` badge, the `location`, the `timeAgo` and the `Purchased —` line, and present the content as illustrative. Better: gate the section behind a flag until real testimonials land.

src/app/api/revalidate/route.ts
```diff
@@ -0,0 +1,13 @@
+import { revalidateTag } from "next/cache";
+import { NextRequest, NextResponse } from "next/server";
+
+export async function POST(req: NextRequest) {
+  const secret = req.nextUrl.searchParams.get("secret");
+
+  if (secret !== process.env.SHOPIFY_REVALIDATE_SECRET) {
```
KatlehoMadaba, line 7: **Blocking — empty-string secret bypasses this check.**

`.env.example` ships `SHOPIFY_REVALIDATE_SECRET=` (blank). If that line is copied into a real `.env`, or the var is created empty in Netlify's UI, `process.env.SHOPIFY_REVALIDATE_SECRET` is `""` — and a request to `/api/revalidate?secret=` gives `"" !== ""` → false → **200, cache purged**. Anyone can then hammer the endpoint and force repeated origin fetches against the Storefront API.

The `.env.example` comment ("Unset means every revalidation request is 401'd") only holds when the var is genuinely *undefined*, not defined-but-empty.

```ts
const expected = process.env.SHOPIFY_REVALIDATE_SECRET;
if (!expected || secret !== expected) {
  return NextResponse.json({ message: "Invalid secret" }, { status: 401 });
}
```

Two further points on this route:

- **Shopify webhooks should be verified by HMAC** (`X-Shopify-Hmac-Sha256` over the raw body with the app secret), not a shared secret in a query string. Query strings land in CDN and Netlify access logs, referrer headers, and browser history — this secret will leak eventually.
- The comparison isn't timing-safe. Minor given the impact ceiling is a cache purge, but `crypto.timingSafeEqual` is a one-liner.

(FWIW the `revalidateTag("products", "max")` two-arg call below is correct — the profile param is required in Next 16.2's type defs.)

src/lib/shopify.ts
```diff
@@ -54,14 +54,33 @@ export interface ShopifyProduct {
...
+  { id: "product:5", name: "Strength & Radiance", category: "Performance", price: 58, originalPrice: 72, image: "", tags: ["Strength", "Glow", "Vitality"], placeholder: true },
+];
+
 export async function getFeaturedProducts(): Promise<ShopifyProduct[]> {
   const domain = process.env.NEXT_PUBLIC_SHOPIFY_STORE_DOMAIN;
   const token = process.env.NEXT_PUBLIC_SHOPIFY_STOREFRONT_TOKEN;
 
   if (!domain || !token) {
-    return [];
+    return placeholderProducts;
```
KatlehoMadaba, line 83: **Blocking — placeholder catalog is now the production fallback.**

This replaces the previous `return []`, so a missing or expired `NEXT_PUBLIC_SHOPIFY_STORE_DOMAIN` / `_TOKEN` on Netlify silently ships a **fake storefront** — five hardcoded products with real-looking prices (R54 "Glow Collagen Blend") and sale badges — instead of failing visibly.

`placeholder: true` only suppresses the image and the cart/wishlist controls. The **prices and "Sale" badges still render**, so the failure mode is a store that looks real and sells nothing.

This also runs against the direction of `master`'s own history (`refactor: drop the placeholder product catalog`). Return `[]` and let `FeaturedProducts` render an empty/error state, or throw at module load when the env vars are absent in production.

Separately: since `getFeaturedProducts` is only called from a server component, the `NEXT_PUBLIC_` prefix on these two vars is unnecessary — it inlines the token into the client bundle for no reason. Storefront tokens are designed to be public, so this isn't urgent, but `SHOPIFY_STORE_DOMAIN` / `SHOPIFY_STOREFRONT_TOKEN` would be tighter.

src/components/sections/Hero.tsx
```diff
@@ -1,14 +1,36 @@
...
+  onVideoReady?: () => void;
+  /** When false, entrance animations hold at "hidden" until flipped to true. */
+  startEntrance?: boolean;
+}
+
+export default function Hero({ onVideoReady, startEntrance = true }: HeroProps) {
+  // Hoists <link rel="preload" as="video"> into the head during SSR so the
+  // browser starts the download with the initial HTML, before hydration.
+  // High priority: the page reveal blocks on this video.
+  preload(HERO_VIDEO_SRC, {
```
KatlehoMadaba, line 22: **Perf — the whole home page is blocked on a 3.7 MB video.**

`HomeHero` covers the page with an opaque loader until `loadeddata` fires on `/video/hero-video.mp4` (3.7 MB), with a 5s fallback. This preload requests it at `fetchPriority: "high"` on top of that.

Two problems:

1. On a slow connection this is a guaranteed **5-second blank brand screen** before any content. LCP will be poor, and Core Web Vitals feed into SEO ranking.
2. **`<link rel="preload" as="video">` is not a supported destination in Chromium** — it's ignored with a console warning, and in some configurations causes a *duplicate* fetch when the `<video>` element then requests the same URL. Worth confirming in DevTools that you get one request, not two.

At minimum, drop `FALLBACK_MS` to ~2s and add a poster image so the hero is meaningful before the video arrives.

src/components/sections/HomeHero.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const reducedMotion = useReducedMotion();
+
+  // Release fn for the first-visit scroll-lock / inert. Held in a ref so the
+  // loader's exit-complete (below) can release it *after* the slide-up finishes,
+  // keeping the page locked while the loader animates away.
+  const releaseLockRef = useRef<(() => void) | null>(null);
+
+  const handleVideoReady = useCallback(() => setVideoReady(true), []);
+
+  useEffect(() => {
```
KatlehoMadaba, line 44: **Bug — blank-hero flash on every repeat visit / reload.**

The pre-paint `skipScript` hides the SSR'd loader via injected CSS when the session flag is set. But `skipped` is initialised from `shownThisSpaSession`, which is `false` after any **full page reload**. So on the first render of a repeat visit:

- the loader is `display:none` (the script hid it), **and**
- `revealed` is `false`, so `startEntrance={false}` renders `Hero` at its `hidden` variant (opacity 0).

The page is blank until this `useEffect` runs and flips `skipped` — and `useEffect` runs *after* paint, so returning visitors should see a flash of empty hero on every reload.

Switching this to `useLayoutEffect` (SSR-guarded) closes the window, since it commits before the browser paints.

Oreginal:
Invalid

PR #48, opened 2026-07-13
Kat/feat/4/shopping cart func

KatlehoMadaba:


Oreginal:
## Code review — Shopify storefront wiring

Reviewed the full diff. Overall this is solid, well-commented work: good use of a GraphQL fragment, graceful fallback when the store isn't configured, three near-duplicate card implementations consolidated into one `ShopifyProductCard`, and it fixes the old sale-badge bug (`compareAmount > price` instead of `> 0`). A few things to address before merge, most important first.

### 🟡 1. Price formatting is inconsistent between the card and the detail page
`src/components/ShopifyProductCard.tsx` defines its **own** `ZAR` formatter with `minimumFractionDigits: 0`, while `src/app/store/[slug]/ProductDetailContent.tsx` uses `formatPrice` from `src/lib/format.ts` (`minimumFractionDigits: 2`).

Result: the same product renders as **`R54`** on the store/featured grid but **`R54,00`** on its detail page. It's also a duplicated formatter. Import `formatPrice` (or add a shared whole-rand variant to `@/lib/format`) and use it in both places — pick one decimal convention.

### 🟡 2. Revalidation webhook: shared secret in the query string, no HMAC
`src/app/api/revalidate/route.ts` reads the secret from `?secret=…`. Query strings leak into server/proxy access logs and `Referer` headers. In order of preference:
- Shopify already signs webhooks with `X-Shopify-Hmac-Sha256` — verifying that HMAC against the raw body is the idiomatic, secure approach and removes the shared secret from the URL entirely.
- At minimum, move the secret to a request header.

The `secret !== process.env…` comparison is also not constant-time (minor for a webhook secret). Fail-closed behavior when the env var is unset is correct and documented — just note that prod **must** set `SHOPIFY_REVALIDATE_SECRET` or on-demand revalidation 401s and only the hourly fallback refreshes.

### 🟢 3. Verify `revalidateTag("products", "max")` actually busts the fetch cache
The two-arg call is correct for Next 16.2.6 (`revalidateTag(tag, profile)`, and `"max"` is a valid cacheLife profile — confirmed against installed types). But the data is tagged via `fetch(..., { next: { tags: ["products"] } })` and the project does **not** enable `cacheComponents`. Worth a quick end-to-end test on a deploy: fire the webhook and confirm the grid updates before the 1h fallback window. Not asserting it's broken — just the one thing that can't be confirmed statically.

### 🟢 4. Minor edge case in the detail-page compare-at price
`showCompareAt = originalPrice != null && price === product.price` in `ProductDetailContent.tsx`. If the cheapest variant is out of stock, `defaultVariant` selects a pricier in-stock one, so `price !== product.price` and the strikethrough sale price silently disappears for a discounted product. Acceptable given the documented "compareAt is a product-level range" limitation, but worth a comment or follow-up.

---

### ⚠️ Please retarget this PR to `main`
This PR currently targets **`master`**, which contradicts the branching rule the PR itself documents in `CLAUDE.md` (feature → `main` → `master`). `master` is now protected by a required **"Only main may PR into master"** status check, so **this PR cannot merge into `master` as-is** — the check will fail for any source branch other than `main`.

Please change the base to **`main`** (Edit → base branch in the PR UI, or `gh pr edit 48 --base main`). Once it's merged and verified on `main`, promotion to production happens via a separate `main → master` PR.

_Findings 1–2 are the blocking ones; 3–4 are worth a look but can be follow-ups._

Oreginal:
Superseded by #62.

This branch had gone 21 commits behind `main` with three merge conflicts, so rather than rebasing in
place I branched off current `main` and **merged this branch in** — your five commits and their
authorship are preserved in #62's history.

All four review findings are addressed there, plus one the review got wrong:
`revalidateTag("products", "max")` doesn't actually hard-invalidate (`"max"` sets expiry a year out,
so only the stale path fires — verified against a running build, the price stayed stale across four
requests). It's now `{ expire: 0 }`.

The one that would have bitten us silently: `ShopifyProductCard.tsx` **auto-merged cleanly** but was
wrong afterwards — it became the shared card for both grids while carrying only the homepage's 5-col
`sizes` string, and the store grid's `priority` eager-load from #50 vanished with no compile error.
Fixed with a per-grid `variant` prop, and `check-images.mjs` now guards it.

Your `compareAmount > price` fix is called out in the PR body so it doesn't get mistaken for a
regression — it's a genuine improvement over what was on `main`.

Closing this one; review is welcome on #62.

PR #49, opened 2026-07-13
Enforce email verification at checkout (#22)

### LINKED ISSUE #22: Email verification is not enforced for protected routes (advisory only)

Oreginal:
## What & why

Closes #22. After signup an account + session exist even though the email is unverified, and nothing blocked an unverified user from reaching `/checkout` — so the "reached checkout ⇒ verified email" invariant future order/payment logic will rely on was unenforced.

**Decision (per issue options): enforce at `/checkout` only.** `/account` stays advisory (keeps its soft "verify your email" banner).

## Approach

Enforcement is **client-side** — the edge `proxy.ts` only sees the presence cookie and can't read Firebase's `emailVerified`, so the gate lives in `CheckoutContent`, matching the existing stale-session guard pattern.

- **`CheckoutContent`** — unverified users are redirected to `/verify-email?from=<full dest>`. The full path+query is preserved (so a future `/checkout?coupon=X` round-trips), routed through the shared open-redirect guard.
- **`AuthContext.reloadUser()`** (new) — Firebase's `emailVerified` isn't live; this reloads the user and returns the fresh value. Memoized (`useCallback`), resilient (`try/catch` → `false`, never throws), and coalesces concurrent calls via an in-flight ref.
- **`VerifyEmailContent`** — polls `reloadUser()` (`VERIFY_POLL_INTERVAL_MS`) + re-checks on tab focus/visibility, then returns the user to their original destination; CTA relabels to "Continue to Checkout". Poll tears down on unmount / sign-out / success.
- **`src/lib/redirect.ts`** (new) — extracted `resolveDestination` out of `LoginContent` so login and verify-email share one open-redirect-safe helper.
- Redirect-once `useRef` guards prevent React Strict Mode from firing double navigations.
- `proxy.ts` comment + `CLAUDE.md` updated to document the decision.

## Verification

- `npm run build` and `tsc --noEmit` pass.
- Browser-tested: verified user → `/checkout` renders; unverified → `/checkout` redirects to `/verify-email?from=%2Fcheckout` with "Continue to Checkout" CTA; `/account` reachable with advisory banner while unverified; poll loop stable with no console errors over multiple cycles.
- Not exercised live (needs Firebase console to flip `emailVerified`): the auto-redirect at the instant verification lands — its polling mechanism is confirmed running cleanly.

### DIFF
--- a/src/app/(auth)/verify-email/VerifyEmailContent.tsx
+++ b/src/app/(auth)/verify-email/VerifyEmailContent.tsx
@@ -1,26 +1,80 @@
 "use client";
 
-import { useEffect, useState } from "react";
+import { Suspense, useCallback, useEffect, useRef, useState } from "react";
 import Link from "next/link";
-import { useRouter } from "next/navigation";
+import { useRouter, useSearchParams } from "next/navigation";
 import { motion, AnimatePresence } from "framer-motion";
 import { Mail, CheckCircle } from "lucide-react";
 import { fadeUp } from "@/lib/motion";
 import { useAuth } from "@/context/AuthContext";
+import { resolveDestination } from "@/lib/redirect";
 import AuthForm from "@/components/auth/AuthForm";
 import AuthErrorToast, { getAuthErrorMessage } from "@/components/auth/AuthErrorToast";
 
-export default function VerifyEmailContent() {
+// How often to re-check emailVerified while the page is open. Firebase has no
+// live listener for verification, so we poll (and also re-check on tab focus).
+const VERIFY_POLL_INTERVAL_MS = 3000;
+
+function VerifyEmail() {
   const router = useRouter();
-  const { user, loading, sendVerification, signOut } = useAuth();
+  const params = useSearchParams();
+  // Where to send the user once verified — the checkout guard passes ?from=/checkout…;
+  // resolveDestination falls back to /account and blocks open redirects.
+  const dest = resolveDestination(params.get("from"));
+  const { user, loading, sendVerification, reloadUser, signOut } = useAuth();
   const [resending, setResending] = useState(false);
   const [resent, setResent] = useState(false);
   const [error, setError] = useState("");
+  // One-shot navigation guard — Strict Mode double-invokes effects, and the
+  // mount check / poll / focus handler could otherwise each fire a navigation.
+  const redirected = useRef(false);
+
+  const navOnce = useCallback(
+    (path: string) => {
+      if (redirected.current) return;
+      redirected.current = true;
+      router.replace(path);
+    },
+    [router]
+  );
 
   // No session (e.g. opened directly) — nothing to verify, send to login.
   useEffect(() => {
-    if (!loading && !user) router.replace("/login");
-  }, [loading, user, router]);
+    if (!loading && !user) navOnce("/login");
+  }, [loading, user, navOnce]);
+
+  // Auto-detect verification: already-verified on mount, a background poll, or
+  // the user returning to this tab after clicking the email link. Because
+  // emailVerified isn't live, each trigger calls reloadUser() (which coalesces
+  // concurrent calls and never throws). Tearing down on unmount / sign-out /
+  // success means no reload requests linger after the page is inactive.
+  useEffect(() => {
+    if (loading || !user) return;
+    if (user.emailVerified) {
+      navOnce(dest);
+      return;
+    }
+
+    let cancelled = false;
+    const check = async () => {
+      const verified = await reloadUser();
+      if (!cancelled && verified) navOnce(dest);
+    };
+
+    const interval = setInterval(check, VERIFY_POLL_INTERVAL_MS);
+    const onVisible = () => {
+      if (document.visibilityState === "visible") check();
+    };
+    document.addEventListener("visibilitychange", onVisible);
+    window.addEventListener("focus", check);
+
+    return () => {
+      cancelled = true;
+      clearInterval(interval);
+      document.removeEventListener("visibilitychange", onVisible);
+      window.removeEventListener("focus", check);
+    };
+  }, [loading, user, reloadUser, navOnce, dest]);
 
   const handleResend = async () => {
     if (resending) return;
@@ -132,7 +186,7 @@ export default function VerifyEmailContent() {
           <motion.button
             variants={fadeUp}
             type="button"
-            onClick={() => router.push("/account")}
+            onClick={() => router.push(dest)}
             whileHover={{ scale: 1.015 }}
             whileTap={{ scale: 0.98 }}
             className="w-full py-4 rounded-full text-white font-semibold text-[11px] tracking-[0.13em] uppercase"
@@ -142,7 +196,9 @@ export default function VerifyEmailContent() {
               boxShadow: "0 6px 24px rgba(201,151,122,0.38)",
             }}
           >
-            Continue to My Account
+            {dest.startsWith("/checkout")
+              ? "Continue to Checkout"
+              : "Continue to My Account"}
           </motion.button>
 
           <motion.button
@@ -159,5 +215,14 @@ export default function VerifyEmailContent() {
 
       <AuthErrorToast message={error} onDismiss={() => setError("")} />
     </>
+  );
+}
+
+export default function VerifyEmailContent() {
+  // useSearchParams requires a Suspense boundary (matches LoginContent).
+  return (
+    <Suspense fallback={null}>
+      <VerifyEmail />
+    </Suspense>
   );
 }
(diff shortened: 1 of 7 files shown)


src/app/checkout/CheckoutContent.tsx
```diff
@@ -15,13 +15,27 @@ export default function CheckoutContent() {
...
+  // cookie): (1) stale-cookie / dead session → login; (2) unverified email →
+  // verify-email, preserving the full destination (incl. query, e.g. a coupon)
+  // so the user lands back here once verified. See #22.
   useEffect(() => {
-    if (!loading && !user) router.replace("/login?from=checkout");
+    if (loading || redirected.current) return;
+    if (!user) {
+      redirected.current = true;
+      router.replace("/login?from=checkout");
+    } else if (!user.emailVerified) {
```
KatlehoMadaba, line 31: This gate is correct for the client-side UX, but it's bypassable by design (disable JS / hit the order API directly / edit client state). The PR says future order/payment logic will *rely* on the "reached checkout ⇒ verified" invariant — that reliance has to be enforced **server-side**, not here: verify `email_verified` from the Firebase ID token on the order/payment endpoint, or in Firestore rules (`request.auth.token.email_verified == true`). Nothing to change in this PR (there's no order surface yet); flagging so the client redirect isn't later mistaken for the enforcement boundary. Worth a note in #22.

src/app/(auth)/verify-email/VerifyEmailContent.tsx
```diff
@@ -142,7 +196,9 @@ export default function VerifyEmailContent() {
               boxShadow: "0 6px 24px rgba(201,151,122,0.38)",
             }}
           >
-            Continue to My Account
+            {dest.startsWith("/checkout")
```
KatlehoMadaba: Because `/checkout` is enforced (unlike `/account`, which is advisory), an unverified user who clicks "Continue to Checkout" gets `router.push("/checkout")` → CheckoutContent's guard fires → bounced straight back here. So on the checkout path this button is a dead loop for the exact state the page is shown in; the auto-redirect already handles the verified case. Consider hiding or disabling it when `dest.startsWith("/checkout")`, rather than relabeling it to something that doesn't work until verification lands.

KatlehoMadaba (commented):
## Review — enforce email verification at checkout (#22)

Clean, well-reasoned PR that matches the option chosen in the issue (enforce at `/checkout`, keep `/account` advisory). The tricky Firebase details are handled correctly:

- **`reloadUser()` + `setTick` is the right pattern.** `currentUser.reload()` mutates the existing `User` object in place and does *not* fire `onAuthStateChanged`, so the React `user` reference never changes — bumping a tick to force the re-read is exactly what's needed, and returning the boolean directly means callers don't even depend on that re-render. I traced the full round-trip: verify → `navOnce(dest)` → `/checkout` reads the now-mutated `user.emailVerified` → gate passes. It works end to end.
- **In-flight coalescing** via the ref is correct — the assignment happens synchronously after the async IIFE yields at its first `await`, so poll + focus can't fire concurrent `reload()`s.
- Strict-Mode one-shot `redirected` refs, the shared `resolveDestination` extraction with the open-redirect guard, and the full-destination round-trip (`?from=/checkout…` preserving query) are all good.

No blocking bugs. One important caveat to carry forward, one UX dead-end, one minor:

### 1. The gate is client-side — it's UX, not a security boundary (carry forward)

This is correct *as scoped* — the edge proxy genuinely can't read `emailVerified`, and the issue chose client-side enforcement. But the PR description frames it as establishing an invariant that *"future order/payment logic will rely on."* A client-side redirect is trivially bypassable (disable JS, call the API directly, edit state), so **that future logic must independently verify `email_verified` server-side** — from the Firebase ID token on the order/payment endpoint, or in Firestore security rules (`request.auth.token.email_verified == true`). There's no order/payment surface in the repo yet, so this isn't a defect here; I'm flagging it so nobody downstream mistakes the checkout redirect for the enforcement point. Worth a line in #22 or the follow-up so it isn't lost. Inline note on the gate.

### 2. "Continue to Checkout" CTA bounces unverified users (UX)

The verify-email page's primary button now relabels to "Continue to Checkout" and does `router.push(dest)`. But `/checkout` is *enforced*, unlike `/account` which is advisory — so an unverified user who clicks it is immediately redirected right back to `/verify-email`. The original "Continue to My Account" worked because account is advisory; the relabeled version is a dead loop for exactly the state the page is shown in. Since verification already auto-redirects the moment it lands, this manual button adds nothing on the checkout path and actively confuses. Consider hiding/disabling it when `dest` is checkout, or dropping it there entirely. Inline note.

### 3. `reloadUser` error spam on a dead session (minor)

If the session is expired/revoked, `reload()` throws every tick and the poll `console.error`s every 3s indefinitely (until `onAuthStateChanged(null)` eventually fires the mount guard to `/login`). Harmless but noisy. Not worth holding the PR — just noting it if you add poll backoff later.

Happy to approve; #1 is documentation/follow-up and #2 is a small tweak.

src/app/checkout/CheckoutContent.tsx
Oreginal: Agreed on all of it — the client gate is UX, and the "reached checkout ⇒ verified" invariant has to be re-enforced at the real trust boundary (ID-token check on the order/payment endpoint and/or `request.auth.token.email_verified == true` in Firestore rules). No order/payment surface exists yet, so nothing to change in this PR.

Since #22 closes with this PR, I've carried it forward as a durable tracking issue — **#57** — rather than a comment here that would die with the issue, so it's on record before any payment logic is written and the client redirect isn't mistaken for the enforcement point.

src/app/(auth)/verify-email/VerifyEmailContent.tsx
Oreginal: Fixed in `732c68f`. You're right it was a dead loop — `router.push("/checkout")` → checkout guard → straight back here, for the exact unverified state the page is shown in.

Rather than relabel or disable, the manual CTA now renders **only for the advisory `/account` destination**; on the checkout path the page relies solely on the auto-redirect (poll + tab-focus), which already fires the instant verification lands. Confirmed `dest` is only ever `/checkout` or `/account` — the checkout guard sets `?from=/checkout…`, and account/signup pass no param → `/account` — so the retained "Continue to My Account" label is accurate for every case the button still shows in. `tsc --noEmit` and `npm run build` both clean.

Oreginal:
@KatlehoMadaba thanks — the Firebase trace (reload-mutates-in-place, no `onAuthStateChanged`, tick-to-re-read) matches exactly how it was reasoned. Addressed all three:

**1. Client gate ≠ security boundary** — fully agreed. Carried forward as **#57**: future order/payment logic must verify `email_verified` server-side (ID token on the endpoint and/or `request.auth.token.email_verified == true` in Firestore rules). Opened as its own issue rather than a note on #22, since #22 closes with this PR and the note would go with it.

**2. "Continue to Checkout" dead loop** — fixed in `732c68f`. The manual CTA now renders only for the advisory `/account` destination; on the checkout path the page relies on the auto-redirect (which already fires the instant verification lands). `dest` is only ever `/checkout` or `/account`, so the retained "Continue to My Account" label stays accurate. Build + `tsc` clean.

**3. `reloadUser` error spam on a dead session** — real, and left as-is per your call (not worth holding the PR). Captured on **#57** as a small follow-up so it's tracked: swallow-or-backoff the `reload()` rejection on the poll path so an expired session doesn't `console.error` every 3s until `onAuthStateChanged(null)` lands. Trivial and can ride with the backoff work if/when it happens.

So: #1 → #57, #2 → code (done), #3 → noted on #57. Clear for another look.

KatlehoMadaba (approved):

Issue #45, opened 2026-07-11
perf: downsize loader logo asset (2956x2817 rendered at 140-180px)


Oreginal:
## Context

Raised in [PR #41](https://github.com/KatOre-Solutions/HeyBeautiful/pull/41) code review by @KatlehoMadaba.

`PageLoader` renders the brand logo with `priority` (it's the loader's centerpiece, so it must
be eagerly loaded and decoded before first paint):

```tsx
<Image
  src="/images/logo-hearts-white.png"
  alt=""
  width={2956}
  height={2817}
  priority
  sizes="(min-width: 768px) 180px, 140px"
  className="w-[140px] md:w-[180px] h-auto select-none"
/>
```

The source asset is **2956×2817** (~100 KB), but it only ever renders at **140–180px**. Next's
`<Image>` will generate appropriately-sized responsive variants at request time, but the
*original* asset is still what's fetched/decoded server-side and cached — a pre-downsized source
would shrink that `priority` (render-blocking) download further.

## Task

- Downsize `public/images/logo-hearts-white.png` to a source resolution that comfortably covers
  the largest rendered size at typical device pixel ratios (e.g. ~2x of 180px → ~360-400px,
  or up to ~2x for retina, whichever the design wants headroom for).
- Confirm visual quality holds at both the 140px and 180px breakpoints, including on retina
  displays.
- Re-run through `next/image` to confirm optimized output size drops accordingly.

## Why it matters

This asset loads with `priority` on every first-visit home page load (it's the full-screen
loader), so its size directly affects loader paint time on slow connections.

Oreginal:
Catered for in #13 / PR #50 — the product photography review swept every source asset in `public/images/`, and this one was already on the list independently. Closing.

**What shipped:**

`logo-hearts-white.png` **2956×2817 → 768×732**, **97.6KB → 23.4KB** (−76%). Ratio preserved (1.0493 → 1.0492), so the `h-auto` layout box is unchanged. The `PageLoader` intrinsic props were updated to match:

```tsx
width={768}   // was 2956
height={732}  // was 2817
```

768px covers the 180px render at DPR4 — generous headroom over the ~360–400px this issue suggested, but it costs little on a flat two-tone PNG and leaves room if the loader ever grows.

`priority` is **retained** deliberately: your reasoning holds, it's the loader's centerpiece and must decode before first paint. This only shrinks what that render-blocking fetch has to pull.

[...]

PR #50, opened 2026-07-15
Product photography review: consistent 4:5 sizing, compression, responsive delivery (#13)

### LINKED ISSUE #13: Product Photography Review
## Description
Review and optimize all product imagery.
## Acceptance Criteria
- [ ] Consistent image sizing
- [ ] Compression
- [ ] Responsive delivery

Oreginal:
Closes #13.

Delivers the three acceptance criteria. The site is server-rendered, so `/_next/image` was already transcoding to AVIF and generating srcsets — the defects were in **how the API was called**, not whether it was available.

## Headline numbers

| | before | after |
|---|---|---|
| Source assets on disk | 3521 KB | **1582 KB** (−55%) |
| A 32px avatar (DPR2) | 31.8 KB | **0.96 KB** |
| `model.jpeg` | 584 KB | **180 KB** |
| `/store` total image transfer | — | 19 requests, 118 KB |

## Consistent sizing

Cards were `aspect-[3/4]` while the detail page was `aspect-[4/5]`, so the same asset cropped differently between views. Everything product-related is now 4:5, declaratively.

The `item-*` sources were **1600×1066 landscape** forced into portrait frames. They're re-cropped to 853×1066. The crop window (x=374..1226) is *exactly* what the 4:5 detail page already showed via `object-cover`, so **the detail view is pixel-identical**; the card, which clipped to x=400..1200 at 3:4, gains a little room.

Also unified the detail thumbnails (5:6) and the cart/wishlist/checkout minis, which were three different pixel pairs within 2% of 4:5. That surfaced a real defect: `w-12 h-15` in `CheckoutContent` — **`h-15` is not a Tailwind class** (the scale skips 15). It was dead, and an inline style was silently doing the work.

## Compression

`scripts/normalize-images.mjs` is committed rather than run ad-hoc, because the outputs are binaries a reviewer must be able to regenerate when new artwork lands. Dry-run by default, idempotent on geometry, **never wired into `build`**.

Every non-product crop is derived from what `object-cover` already centres on, so they are **visually lossless by construction** — the browser was discarding exactly those pixels already.

`logo.jpeg` deliberately **stays JPEG**: it has no alpha (the circle is CSS `rounded-full`) and as PNG it would be several times larger. Only its resolution was wasteful. `sub-logo-white.png` is deleted — verified unreferenced across the entire project, with no dynamic path construction anywhere.

## Responsive delivery

- **8 `fill` images had no `sizes`**, so Next defaulted to `100vw` and a 32px avatar fetched a ~1920px candidate.
- **4 had a *wrong* `sizes`.** `BrandStory` claimed `50vw` from 769px up, but its grid is `lg:grid-cols-2` — single-column until 1024. **Every tablet was rendering it at roughly half the resolution it needed.** That was a live bug on `main`, not just waste.
- Every trailing px value now carries its derivation from the Tailwind layout inline, so a future grid change can be recomputed rather than guessed. Verified against the rendered box: predicted 246px, measured 245px.
- The store grid's first row is now eager (`priority={index < 4}`) — it's the LCP. `FeaturedProducts` deliberately gets **none**: it sits below a `<video>` hero that already preloads itself, so preloading there would contend with the real LCP.
- **Removed `priority` from the `(auth)` brand panel.** It's `hidden lg:flex`, but `priority` emits an unconditional `<link rel=preload>` — mobile was downloading a decorative image for a `display:none` element. Verified: 0 preloads on `/login` now, was 1.
- Documented `qualities: [75]` in `next.config.ts`. Next 16 rejects any other `quality` with a **400 at request time, never at build time** — a `quality={90}` would pass typecheck, lint and `next build`, then break in production.

## Visual changes: intended vs not

The intended visual changes are both product-framing unifications: cards 3:4 → 4:5, and the `BundleCard` stacked thumbnails square (1:1) → 4:5 to match (they're product shots, so they shouldn't be the one place the old shape survives). That same `BundleCard` edit also adds `relative` to the thumbnail container, fixing a latent bug where `<Image fill>` had no positioned parent. Everything else is pixel-identical by construction. Product margins after cropping are symmetric to within 16px and nothing clips.

## Regression guard

`scripts/check-images.mjs` (zero-dep, `npm run images:check`) asserts every `fill` Image has `sizes`, the product cards stay 4:5, and only the intended cards get `priority`. **Each of the four assertions was verified to actually fail when deliberately broken** — a green check that can't go red is worthless. It's regex over source, not an AST pass: it catches the realistic regression and can't produce a false alarm that blocks a good build.

⚠️ **It is not wired into `npm run lint`, because `npm run lint` is already broken on `main`** — Next 16 removed the `next lint` subcommand, and there's no ESLint config file, no CI, and no git hooks. Chaining the guard behind a permanently-red command would orphan it. Fixing lint means adding an ESLint config: a separate PR. See the notes below.

## Verification

- `npm run build` from the correctly-cased path — clean, zero `InvariantError`.
- srcset ladder confirmed 16w→2048w; AVIF bytes increase monotonically (3.0/4.9/8.5/11.4 KB at w=256/384/640/1200).
- `w=1200` returns 11.4KB from an 853px source, confirming `withoutEnlargement` never upscales.
- `q=90` → **400** `"q" parameter (quality) of 90 is not allowed`; `q=75` → 200.
- Browser: all 5 cards measure **ratio 0.800 exactly**; `item-1..4` eager, card 5 lazy; zero horizontal overflow.

## Out of scope (tracked separately)

The artwork has defects code can't fix — packaging label mismatches and a **"22g Plant Protien"** typo printed on four products. Filed as a separate issue for a designer. `ProductCard`/`ShopifyProductCard` deduplication and blur placeholders are deferred per the plan.


Oreginal:
Cross-linking: the reason `scripts/check-images.mjs` ships standalone rather than chained into `npm run lint` is now tracked as #52 — `next lint` was removed in Next 16 and there's no ESLint config, so lint is permanently red on `main`. Wiring the guard in is a one-liner once that lands.

The plan deviations (the bbox-vs-centred-crop correction, the unachievable 1200×1500 target, and this) are written up on #13.

Oreginal:
### Also covers #45 (closed)

This PR includes the loader-logo downsize that #45 asked for — it was raised separately in PR #41's review by @KatlehoMadaba, and the asset sweep here happened to cover it.

`logo-hearts-white.png` **2956×2817 → 768×732**, **97.6KB → 23.4KB** (−76%), ratio preserved so the `h-auto` layout box is unchanged. `PageLoader`'s `width`/`height` props updated to match. `priority` retained — it's the loader's centerpiece and must decode before first paint; this just shrinks what that render-blocking fetch pulls.

#45 has been closed with the detail rather than auto-closed from this PR body, since it predates this work and deserves its own record.

**Worth flagging for reviewers,** because it applies to how we reason about assets generally: #45's premise was that the oversized original "is still what's fetched/decoded server-side and cached". Not quite — `/_next/image` serves a resized AVIF/WebP variant, so that 100KB PNG never hit the wire. It *was* the optimizer's input, and a cold-cache transcode of a 2956px PNG is real work, so the win is real — just smaller than the raw KB delta implies. Same caveat applies to the source-asset numbers in this PR's summary: **3521KB → 1582KB is optimizer input and repo weight, not bytes off the wire.** The wire numbers are the separate ones (avatar 31.8KB → 0.96KB, `/store` at 118KB total), and those come from the `sizes` fixes rather than the re-encoding.

KatlehoMadaba (commented):
Automated review (high effort). Strong PR — I traced every pre-crop ratio in `normalize-images.mjs` against its display container and they all match the `object-cover` aspect (`model`→4:5 / BrandStory, `model-2`→3:4 / Lifestyle, `story-3`→3:4, `story-4`→4:3 / Testimonials; multi-ratio assets like `inspo-*` and stories 1–2 are correctly re-encode-only), so the visually-lossless claim holds. The BrandStory `50vw`-from-769px fix is a real tablet under-fetch (`lg:grid-cols-2` is single-column until 1024), and the grid shapes behind the new `sizes` values check out. No correctness bugs found. A few low-severity notes inline.

scripts/check-images.mjs
```diff
@@ -0,0 +1,112 @@
+// Guards the image-delivery invariants established in issue #13.
+// Runs as part of `npm run lint`, or on its own via `npm run images:check`.
```
KatlehoMadaba: This comment says the guard *"Runs as part of `npm run lint`"*, but it isn't wired in — `package.json`'s `lint` is still just `next lint`, and the PR description itself explains it's deliberately kept out of lint (which is broken on `main`). The comment tells a future maintainer that image-delivery regressions are caught automatically, when in reality nothing runs `images:check` on its own — no lint step, no CI, no hook. Worth softening to match reality (e.g. "run manually via `npm run images:check`") so nobody assumes it's guarding merges.

scripts/check-images.mjs
```diff
@@ -0,0 +1,112 @@
...
+ * would contend with the real LCP.
+ */
+function checkPriority(file, src) {
+  if (/StoreContent\.tsx$/.test(file) && !src.includes(EAGER_STORE_CARDS)) {
+    fail(file, 1, `store grid must eager-load its first row via \`${EAGER_STORE_CARDS}\``);
+  }
+  if (/FeaturedProducts\.tsx$/.test(file)) {
+    const p = src.search(/\bpriority\b/);
+    if (p !== -1) {
+      fail(file, lineAt(src, p), "FeaturedProducts must not set `priority` — the homepage LCP is the hero video");
```
KatlehoMadaba: This assertion is nearly vacuous. `FeaturedProducts.tsx` delegates rendering to `ShopifyProductCard`, which has no `priority` prop at all — so the only way the literal word `priority` appears in `FeaturedProducts.tsx` is a stray comment. It doesn't actually guard "the featured grid must not eager-load": a `priority` added inside `ShopifyProductCard` (or a `priority` prop threaded through it, as `ProductCard` now has) would sail past this check. If the real invariant is "no eager-loading in the homepage featured section," it needs to assert against where `priority` could actually be set.

scripts/normalize-images.mjs
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  console.log(
+    `\n  total ${fmtKB(totalBefore)} -> ${fmtKB(totalAfter)}  (${(
+      (1 - totalAfter / totalBefore) * 100
+    ).toFixed(1)}% smaller)`,
+  );
+
+  if (failed) {
+    console.error("\nFAIL: a crop would clip its product. Re-tune TRIM_THRESHOLD or the target.");
```
KatlehoMadaba: Stale constant name in the failure message: it says *"Re-tune TRIM_THRESHOLD"*, but there is no `TRIM_THRESHOLD` in this file — the knobs that actually govern this failure are `OBJECT_THRESHOLD` (product-vs-shadow detection) and `MAX_CENTRE_SKEW` (the skew tolerance). This message fires exactly when someone is debugging a clipped crop, so pointing them at a nonexistent variable is the worst moment for it.

src/components/BundleCard.tsx
```diff
@@ -140,13 +140,13 @@ export default function BundleCard({
             {bundle.products.map((img, i) => (
               <div
                 key={i}
-                className="w-14 h-14 rounded-xl overflow-hidden border-2 border-white/70 shadow-soft flex-shrink-0"
+                className="relative w-14 aspect-[4/5] rounded-xl overflow-hidden border-2 border-white/70 shadow-soft flex-shrink-0"
```
KatlehoMadaba, line 143: Worth a conscious confirm: this reframes the bundle thumbnails from square (`w-14 h-14`, 56×56) to `w-14 aspect-[4/5]` (56×70) — a real shape change beyond the stated "product cards 3:4 → 4:5" intent. 

Upside worth noting: adding `relative` here also fixes a latent bug — the previous markup used `<Image fill>` with a parent that had no `position`, so the fill image wasn't actually constrained to the 56px box (it resolved against the nearest positioned ancestor). So this line is a genuine fix *and* a reframe; just make sure the taller portrait thumbnails in the rotated stack are intended.

scripts/check-images.mjs
Oreginal: Good catch — it was fully vacuous. Fixed in 2368b2b: the guard now scans the `<Image>` tags of both `FeaturedProducts.tsx` **and** `ShopifyProductCard.tsx` (the only consumer, and where `priority` could actually be set). It treats `priority` / `priority={true}` / `priority={<expr>}` as eager-loading and allows only an explicit `priority={false}` opt-out. Verified it now fails when `priority` is injected into ShopifyProductCard's Image, and passes on `priority={false}`.

scripts/normalize-images.mjs
Oreginal: Fixed in 2368b2b — the message now points at the real knobs, `OBJECT_THRESHOLD` / `MAX_CENTRE_SKEW`, instead of the nonexistent `TRIM_THRESHOLD`.

scripts/check-images.mjs
Oreginal: You're right that it isn't wired in today. Rather than soften the wording now and re-touch it later, this is being handled as part of #52 (fixing the broken `next lint`): that issue chains the guard into lint (`eslint . && node scripts/check-images.mjs`), at which point the comment becomes accurate. Tracking the comment fix on #52's checklist so it lands together.

scripts/check-images.mjs
```diff
@@ -0,0 +1,125 @@
+// Guards the image-delivery invariants established in issue #13.
+// Runs as part of `npm run lint`, or on its own via `npm run images:check`.
```
KatlehoMadaba: This says the check "Runs as part of `npm run lint`", but it doesn't — `package.json`'s `lint` is still `next lint`, and the PR description explains it's deliberately *not* wired into lint (lint is broken on Next 16). The first line a reader sees currently tells them the opposite of what's true. Suggest something like: `// Run via npm run images:check. NOT wired into npm run lint — see PR #50 / issue #13 for why.`

src/components/BundleCard.tsx
KatlehoMadaba, line 143: Two things here:

1. This changes the bundle thumbnails from square (`w-14 h-14`, 1:1) to `aspect-[4/5]` (56×70). That's a visible shape change, and the PR states the *only* intended visual change is the product cards going 3:4 → 4:5. Worth confirming this square→portrait shift is deliberate — it's a different starting ratio than the cards.
2. Adding `relative` is a genuine fix: the previous container had no positioning context, so the `<Image fill>` below was positioning against some ancestor rather than this box. Nice catch (intentional or not).

KatlehoMadaba (commented):
## Review — image delivery normalization (#13)

One of the cleaner PRs I've reviewed on this repo. The work is real and the diagnoses hold up: the `BrandStory` `50vw`-from-769px bug (single-column grid until `lg`) was genuinely serving every tablet a half-resolution image, the `(auth)` panel's `priority` was preloading a `display:none` image on mobile, and the `h-15`-isn't-a-Tailwind-class find is correct (the scale skips 15). I spot-checked the `sizes` derivations against the grids and they're right, including the deliberate use of `767/1023/1279/1503` to sit just under Tailwind's `min-width` breakpoints so the queries don't overlap at the boundary. The `qualities: [75]` note about the request-time-only 400 is a genuinely useful trap to have documented.

Nothing here blocks. A few things worth addressing before merge, none critical:

### 1. `check-images.mjs` header contradicts reality (and the PR body)

The script's opening comment says it *"Runs as part of `npm run lint`."* It doesn't — `lint` is still `next lint`, and the PR description itself explains at length why the guard is deliberately *not* wired into lint (lint is broken on Next 16). So the one line a future reader sees first tells them the opposite of the truth. Fix the comment to match — inline note below.

### 2. The regression guard is manual-only

Following from #1: `images:check` isn't in `lint`, there's no CI, and there are no git hooks — so nothing runs it automatically. It's a good check (I like that you verified each assertion actually fails when broken), but a guard that depends on someone remembering `npm run images:check` tends to rot. That's disclosed in the PR body, so this is just flagging it as *the* limitation of the change rather than a defect. If a lightweight `pre-commit` hook or a one-line CI step is in reach, it's what turns this from documentation into enforcement. Reasonable to defer to the ESLint-config PR you mention.

### 3. Bundle thumbnails go square → 4:5 — intended?

`BundleCard` changes the little rotated product thumbnails from `w-14 h-14` (1:1) to `w-14 aspect-[4/5]` (56×70). That's a visible shape change, and the PR says *"the only intended visual change is product framing (cards 3:4 → 4:5)."* This is arguably the same unification, but it's a different starting ratio (square, not 3:4) and isn't called out — worth a one-line confirmation it's deliberate. Note the same edit also adds `relative` to the container, which the `<Image fill>` actually needed and was missing before, so the thumbnails were likely mispositioned prior to this — a real latent fix, good.

### Nits

- **`sharp` as a direct `devDependency` is correct** — the normalize script only runs locally and Next provides its own sharp for the runtime optimizer, so this doesn't affect production image handling. Just noting it's the right call, not a concern.
- The image binaries themselves can't be verified in review (no pixels visible via the diff). Trust rests on the script logic + your measurements (245 vs 246px predicted, ratio 0.800 exact). That's inherent to the task, not something to fix here.

Happy to approve once the header comment in #1 is corrected; #2 and #3 can be follow-ups or a sentence in the PR.

src/components/BundleCard.tsx
Oreginal: Confirmed deliberate on both counts.

1. **The square → 4:5 reframe is intentional.** These minis are product shots, so they should read at the same aspect as everywhere else product art appears now — the point of this PR is that a product never crops differently between views. Leaving them 1:1 would have made the bundle previews the one place the old shape survived. `w-14 aspect-[4/5]` (56×70) is the same ratio as the cards, just scaled to the thumbnail box. So it's the same unification, applied to a square starting point rather than a 3:4 one — you're right that it wasn't spelled out in the "only intended change" line, and it should have been; treat this as that confirmation.

2. **The `relative` add is the fix you flagged, and it's real.** The previous markup had `<Image fill>` with an unpositioned parent, so the fill resolved against the nearest positioned ancestor instead of the 56px box — the minis were latently mispositioned. Good catch calling it out as a genuine fix rather than incidental.

I'll add a one-line note to the PR body so the square→4:5 shift is on the record as intended rather than implied.

Oreginal:
@KatlehoMadaba header fix is up in `2de0a2a` — the `check-images.mjs` opening comment now says it runs manually via `npm run images:check` and is explicitly **not** wired into `npm run lint`, with a pointer to #13/#52 for why. That's the line you gated approval on.

Rolling up the rest of your review:
- **Vacuous `priority` assertion** and **stale `TRIM_THRESHOLD` message** — both fixed in `2368b2b`.
- **BundleCard square → 4:5** — confirmed deliberate on the inline thread (same product-framing unification, applied to a square start), and the `relative` add is the latent `<Image fill>` fix you flagged. Also noted it in the PR body so it's on the record as intended.
- **Manual-only guard (no CI/hooks)** — deferring to #52 as you suggested; it chains the guard into lint once `next lint` is replaced.

Should be clear for another look whenever you're ready.

KatlehoMadaba (approved):

Issue #52, opened 2026-07-15
Lint is broken: next lint removed in Next 16, no ESLint config exists


Oreginal:
`npm run lint` fails on every branch, including `main`. It has been dead for a while — nothing in the project runs it, so it went unnoticed. Surfaced while working #13 / PR #50.

## Reproduce

```
$ npm run lint
> next lint
Invalid project directory provided, no such directory: .../website/lint
```

Note the error: `next` is treating `lint` as a **directory argument**, because the subcommand no longer exists. `next --help` on 16.2.6 lists `build`, `dev`, `start`, `info`, `telemetry`, `typegen`, `upgrade`, `experimental-analyze`, `experimental-test`, `internal` — **no `lint`**. Next 16 removed it.

## There are actually three problems stacked

**1. `next lint` no longer exists** (Next 16 removed it; the migration is to invoke ESLint directly).

**2. There is no ESLint config file at all.** Bare `eslint` also fails:

```
$ npx eslint src/components/Footer.tsx
ESLint couldn't find an eslint.config.(js|mjs|cjs) file.
From ESLint v9.0.0, the default configuration file is now eslint.config.js.
```

ESLint 9 requires flat config, and there's no `eslint.config.*` and no `.eslintrc.*` anywhere. `next lint` used to generate/manage this implicitly, so removing it left nothing behind.

**3. Version skew.** `eslint-config-next` is pinned at **15.3.2** while `next` is **16.2.6**:

| package | installed | declared |
|---|---|---|
| `next` | 16.2.6 | `^16.2.6` |
| `eslint` | 9.39.4 | `^9` |
| `eslint-config-next` | **15.3.2** | `15.3.2` (exact pin) |
| `@next/eslint-plugin-next` | **15.3.2** | (transitive) |

## Why it matters now

There is no CI, no husky, and no pre-commit hook — `npm run lint` is the only lint entry point, and it is permanently red. That means **no static analysis runs anywhere in this project**.

It also blocked something concrete in PR #50: that PR adds `scripts/check-images.mjs`, a zero-dependency guard asserting every `fill` `<Image>` declares `sizes`, that product cards stay 4:5, and that only the intended cards get `priority`. The natural home is:

```json
"lint": "next lint && node scripts/check-images.mjs"
```

but chaining it behind a permanently-failing command would mean **the guard never runs**. So it currently ships standalone as `npm run images:check`. Once lint works, wiring it in is a one-liner.

## Suggested fix

[...]

PR #53, opened 2026-07-15
Hero video: 69% smaller, poster-based first paint, Save-Data + reduced motion (#11)

### LINKED ISSUE #11: Hero Section Refinement
## Description
Improve video loading and cinematic experience.

## Acceptance Criteria
- [ ] Optimized video
- [ ] Mobile video support
- [ ] Improved loading state
- [ ] Smooth animations

Oreginal:
Closes #11.

## ⚠️ Read this first: the diagnosis changed

I validated the plan's assumptions against `main` before writing code. **Three of four were false**, including the stated root cause. This is an **optimization**, not the bug fix it was scoped as.

| assumption | verdict | evidence |
|---|---|---|
| Browser must download all 3.7MB before frame 1 | **FALSE** | Chrome range-seeks the tail for `moov`. Logged: `bytes=0-100`, `bytes=0-`, **`bytes=3670016-`** ← tail seek (moov lives at 3,694,674), `bytes=32768-`. `loadeddata` fires after **~32KB**. |
| The 5s fallback fires, revealing a black hero | **FALSE** | `loadeddata` at **3.6s** at slow-3G rate — comfortably under `FALLBACK_MS=5000`. That failure mode does not exist. |
| The preload link double-fetches | **FALSE** | Resource Timing: **exactly 1 request**. (The DevTools extension reported 503s curl couldn't reproduce — an artifact.) |
| moov after mdat costs a tail seek | **TRUE** | faststart files request only `bytes=0-`; no tail seek. |

**So faststart is not the headline — the size reduction is.** Controlled experiment, A = original, B = **control** (identical content remuxed `-c copy -movflags +faststart`, isolating faststart from size), C = shipped candidate. Median time to `loadeddata`, 3 runs:

| rate | A (current) | B (faststart only) | C (this PR) |
|---|---|---|---|
| 250 KB/s | 1,031ms | 857ms (−17%) | **337ms (−67%)** |
| 50 KB/s (slow 3G) | 3,602ms | 3,326ms (−8%) | **1,228ms (−66%)** |

Faststart is still included — it's free with the re-encode and removes a real extra round trip — but it's reported as modest. My rig has no network RTT, which *understates* it.

## Measured results

| metric | before | after |
|---|---|---|
| Hero video | 3,701,584 B | **1,129,500 B** (−69%) |
| Poster | — | **43,956 B** (new cost, counted) |
| **Hero media total** | 3,701,584 B | **1,173,456 B (−68%)** |
| `loadeddata` @50KB/s | 3,602ms | **1,228ms (−66%)** |
| Video requests, cold | 1 | **1** (0 under Save-Data / reduced motion) |
| CLS | 0 | **0** |
| Repo weight | +3.7MB dead duplicate | removed |
| SSIM vs master | — | 0.977 |

**LCP: not reported.** The LCP entry never fired in my automated (backgrounded) tab, so I have no trustworthy number. The LCP element is hero *text*, not the video, and the reveal is gated on the video — so it should improve because the gate clears sooner, but I'm not claiming a figure I didn't measure.

## Poster → playback handover

Measured, not eyeballed. Poster **YAVG 133.435** vs the video's frame 0 **133.39** — a 0.03% difference, SSIM **0.981**. No perceptible pop.

This is why the poster is cut from the **encoded** output, not the master: the master's frame 0 is YAVG 134.48, and the clip **darkens 21% over its 8s** (134.48 → 105.71), so any later frame would be visibly darker than the video's opening.

## Browser matrix — actually tested

| browser | `navigator.connection` | result |
|---|---|---|
| **Chrome** | `saveData: false, effectiveType: "4g"` | ✅ measured |
| **Edge** | `saveData: false, effectiveType: "4g"` | ✅ measured (headless) |
| **Firefox** | — | ⚠️ **not installed locally — untested** |
| **Safari** | — | ⚠️ **macOS/iOS only — untested on this Windows host** |

The untested path is the important one, so it's covered by a truth-table test against the real module (9/9 pass), including **`undefined` → loads the video** rather than withholding it.

⚠️ **Reviewer note:** Chrome DevTools throttling may change `effectiveType` to `2g`/`slow-2g`, which fires the skip branch — so "test the video on Slow 3G" can show **no video and just the poster**, which is correct behaviour, not a bug. My throttling was server-side (real byte-rate limiting), which does *not* alter `effectiveType`, so the numbers above measure the video path.

## A regression I introduced and fixed

Removing `autoPlay` meant **nothing retried playback after a background tab suspended it**. Chrome fires `play → playing → pause` within ~12ms in a hidden tab; `autoPlay` used to retry for us. Anyone opening the page via cmd+click would have seen a permanently frozen first frame. Fixed with a `visibilitychange` retry, verified by faking the visibility transition and asserting `play()` is re-invoked.

## What was intentionally left out

Declined, not missed:

- **`h-screen` → `h-svh`** — a real bug (`100vh` overshoots iOS Safari; this hero is bottom-anchored), but serves no AC here and changes mobile layout. → follow-up.
- **`website/images/*` mirror** — same dead-duplicate mistake as `video/`, but #13's territory. → follow-up.
- **Cache headers for `/video/*`** — path isn't content-hashed, so `immutable` is a footgun; Netlify already 304s repeat visits.
- **`FALLBACK_MS`** — measured 1.2s vs a 5000ms timeout; it never fires, and with a poster it degrades to a still frame. Changing it is diff and risk for no visible benefit.
- **AV1 / VP9 / 480p variant** — VP9 measured *strictly worse* than H.264 (1.68MB @ SSIM 0.955 vs 1.13MB @ 0.977). AV1 matches h264-crf26 quality at 37% less, but needs a second asset for ~220KB.
- **Site-wide `MotionConfig`, loader `MIN_MS`** — follow-ups.

## Follow-ups filed

The video carries a visible **"Veo" AI watermark** (pre-existing on `main`, not introduced here), and the 8s loop **pops 27% brighter** on every restart. Both need the designer. See linked issues.

## Verification

- `npm run build` — clean, zero `InvariantError`, 16/16 static pages. (The `Shopify fetch error: 401` in the log is pre-existing — no Shopify token.)
- `npm run video:check` — green. **Every assertion verified to fail when deliberately broken**: faststart lost, audio re-added (with faststart preserved, so it isolates the audio check), poster stale, `autoPlay` re-added, `preload()` retargeted at the video, `poster` removed, `preload="auto"`.
- Poster staleness uses a **content hash**, not mtime — git doesn't store mtimes, so a clone stamps every file with checkout time and a timestamp compare is noise.
- `npm run video:normalize` is idempotent (re-run skips; `--force` overrides) and **adds no npm dependencies** — ffmpeg's bundled `libwebp` encodes the poster.


src/components/sections/Hero.tsx
```diff
@@ -2,11 +2,38 @@
...
+ *
+ * `hidden` is reused by reference and must stay identical to the original:
+ * framer-motion serialises it into the SSR'd style, and useReducedMotion() is
+ * null server-side, so a reduced-motion-dependent `hidden` would mismatch on
+ * hydration and remount the <video> (see HomeHero's note on remounting).
+ * Only `transition` differs — it is runtime-only and never serialised.
+ *
+ * Opacity still fades; it's movement that prefers-reduced-motion is about.
+ */
+const reducedTransition = {
```
KatlehoMadaba, line 24: `reducedTransition` flattens `y` and `filter` instantly, which fully neutralizes `fadeUp` and `heroEntrance`. But `shimmerLine` animates `scaleX: 0 → 1` (a horizontal wipe), and there's no per-property override for it — so `shimmerLineReduced` still plays that scale animation over 0.4s under `prefers-reduced-motion`. Since a scaleX wipe is movement, consider adding `scaleX: { duration: 0 }` here:

```ts
const reducedTransition = {
  duration: 0.4,
  ease: ease.out,
  y: { duration: 0 },
  filter: { duration: 0 },
  scaleX: { duration: 0 },
};
```

The `hidden` state (`scaleX: 0`) is unchanged, so there's no hydration impact — the line just snaps to full width while opacity fades in, instead of wiping.

KatlehoMadaba (commented):
## Review — hero video optimization (#11)

Excellent work, and the intellectual honesty stands out: opening by disproving three of the four scoped assumptions (including the stated root cause) and re-framing this as an optimization rather than a bug fix is exactly the right thing to do, and rare. The controlled A/B/C experiment isolating faststart from the size reduction, and *not* reporting an LCP number you couldn't trust, are both the right calls.

I verified the parts I couldn't take on faith:

- **Hydration safety of the reduced-motion variants holds.** `useReducedMotion` in framer-motion 12 reads `matchMedia` synchronously in a `useState` initializer and never updates, so it's stable from first client render — your `[reducedMotion]` effect genuinely runs once, and preserving `hidden` by reference (only `transition` differs, which isn't serialized) means SSR and client emit the same hidden style. The reasoning in the comments is correct.
- **The dead root-level `video/hero-video.mp4` is gone**, and the only remaining video refs are under `public/video` plus the scripts. Good catch removing the 3.7MB duplicate.
- **`shouldLoadHeroVideo()` degrades correctly** — `navigator.connection` absent (Safari/Firefox) returns `true`, so those untested browsers get the video rather than having it withheld.

Nothing blocks. One real gap and two notes:

### 1. `shimmerLineReduced` still animates under reduced motion (minor)

`reducedTransition` flattens `y` and `filter` to `duration: 0`, which fully covers `fadeUp` and `heroEntrance` (they animate opacity + y + blur). But `shimmerLine` animates **`scaleX: 0 → 1`** — a horizontal wipe — and there's no `scaleX: { duration: 0 }` override, so under `prefers-reduced-motion` that deco line still scales in over 0.4s. A scaleX wipe is movement, which is the thing the feature is meant to suppress. Adding `scaleX: { duration: 0 }` to `reducedTransition` closes it (the `hidden` `scaleX: 0` stays identical, so no hydration impact). Inline note below.

### 2. The guard isn't enforced automatically (recurring)

`video:check` is a good gate — and I like that you verified each assertion fails when deliberately broken — but like `images:check` from #50 it's manual-only: not in `lint` (still `next lint`, dead on Next 16), no CI, no hooks. It also hard-requires ffmpeg/ffprobe on PATH for the audio check, so it can't run in an environment without them. Same follow-up as #50: an ESLint config / CI step is what turns these guards from documentation into enforcement.

### 3. The untested path is the real residual risk (acknowledged)

You've flagged this yourself: Safari/iOS is where `<video>` autoplay, `loadeddata` timing, and WebP posters are all most fragile, and it's untested on your Windows host. The truth-table test covers the `network.ts` decision, but not the actual playback/first-paint handover on WebKit. Worth having someone smoke-test on a real iOS device before this rides to `master` — the `visibilitychange` retry and iOS Low Power Mode `play()` rejection are exactly the cases that don't show up on Chrome.

Happy to approve after #1; #2 and #3 are follow-ups / a verification step rather than code changes.

src/components/sections/Hero.tsx
Oreginal: Fixed in `b13c594` — added `scaleX: { duration: 0 }` to `reducedTransition`, exactly as suggested. Confirmed `shimmerLine` was the only variant animating `scaleX` (`motion.ts:143`, `scaleX: 0 → 1`), so this is the one gap; `fadeUp`/`heroEntrance` were already fully covered by the `y`/`filter` overrides. `hidden.scaleX` stays `0`, so no hydration impact — the deco line now snaps to full width while opacity fades, instead of wiping. `npm run build` still clean (16/16 static).

Oreginal:
@KatlehoMadaba thanks — thorough review. Addressed all three:

**1. `shimmerLineReduced` scaleX wipe** — fixed in `b13c594`, added `scaleX: { duration: 0 }` to `reducedTransition` (replied inline). `npm run build` still clean, 16/16 static. This was the one you gated approval on.

**2. Guard is manual-only** — agreed, and it's the same gap as `images:check` from #50. Rolling both into the ESLint-config / CI follow-up tracked on #52: once `next lint` is replaced, `video:check` and `images:check` chain in together (`eslint . && node scripts/check-images.mjs && node scripts/video-check.mjs`). Good call on the ffmpeg/ffprobe dependency too — in CI that means the video check runs in a job with ffmpeg on PATH, or guards itself to skip-with-warning when the binaries are absent rather than hard-failing. I'll capture that nuance on #52.

**3. Untested WebKit path** — you're right that this is the real residual risk, and it's exactly the fragile surface (autoplay policy, `loadeddata` timing, WebP poster, the new `visibilitychange` retry, iOS Low Power Mode `play()` rejection). Filed as **#56** so it isn't forgotten — a real-iOS smoke-test with a concrete checklist, marked as blocking the `main → master` promotion but not the merge to `main`.

So: #1 is code (done), #2 → #52, #3 → #56. Clear for another look whenever you're ready.

KatlehoMadaba (approved):

PR #73, opened 2026-08-14
feat: rose gold gradient auth with the angled blade transition

Oreginal:
Rebuilds `/login` and `/signup` around the official Hey Beautiful palette, with the signature interaction: an angled blade carrying the editorial artwork sweeps across the card, the auth mode swaps **while hidden beneath it**, and the blade continues to its opposite resting position to reveal the new form.

Two commits: the desktop build, then mobile support.

## The blade

The blade **is** the branded panel — not an overlay flying over a separate brand column. On desktop it rests left for sign-in, right for sign-up, and the form mirrors sides with it, so the swap reads as a genuine left/right flip.

`/login` and `/signup` stay real, separate URLs. They now live in an `(auth)/(card)` route group whose shared client layout App Router never remounts across sibling navigation — that's what lets one blade element animate straight through the route change instead of being torn down mid-sweep. Nothing that links to these routes had to change (`proxy.ts`, `Navbar`, `CartSidebar`, `CheckoutContent`, `AccountContent` all keep working).

## Mobile

The blade originally shipped with **no breakpoint prefixes at all**, so below `lg` it still rendered as a tall diagonal panel covering roughly the left half of the card — sitting on top of the form. The card also set `min-h-[720px]` unconditionally with the form slot absolutely positioned; an absolute child contributes no height, so a signup form taller than the card was **clipped rather than scrolled**.

Below `lg` the blade is now a branded band across the top with a diagonal bottom edge, and the form stacks beneath it in normal flow so the card grows with its content. The interaction is kept, rotated to portrait: **the band dips down to cover the card, the mode swaps while hidden, and it retracts to reveal the incoming form.** Desktop is unchanged.

- `useIsDesktop` — `useSyncExternalStore` + `matchMedia`, so the first client render already knows the axis and cannot flash a wrong-axis park. 1024px is now encoded in both this query and the `lg:` classes; commented on both sides since there is no shared breakpoint token in this project.
- The transition is axis-aware: X in `%` on desktop, Y in **measured px** on mobile, because the card height is content-driven and differs between sign-in and sign-up. Rest is **re-measured before the reveal** so it lands against the incoming pane height, not the outgoing one.
- **Both axes are always written, never just the active one.** Framer composes against the element current transform, so animating `y` with a stale `translateX` still on the blade would park it off-card right after a breakpoint cross.
- `ResizeObserver` re-parks the band when the card resizes without a React render (image decode, font swap, rotation).
- `clip-path` moved to `.auth-blade-clip` in `globals.css` — the shape changes at a breakpoint and an inline style cannot carry one. First `@media` block in the project, hence the comment.
- **Band copy is desktop-only, and `BAND_H` went 200 → 240.** On a short band the copy landed directly on top of the artwork own baked-in HEY BEAUTIFUL lockup *and* duplicated the form heading a few pixels below. Rather than shrink the type, the photograph carries the branding on mobile and the form supplies the words.

## Route-readiness handshake

The reveal never starts on assumed timing. A single `requestAnimationFrame` after `router.replace()` only guarantees a frame boundary, not that the incoming route committed — on a slow device the blade would open over an empty pane.

Instead: `router.replace()` fires at full cover, then phase 3 blocks until the incoming pane calls `notifyPaneReady(mode)` from a mount effect (post-commit, so it is genuinely mounted) wrapped in one rAF for paint.

| Failure | Mitigation |
| --- | --- |
| Pane never mounts | `Promise.race` with a ~1200ms watchdog; reveals anyway rather than stranding the blade over the card |
| Pane mounts before the wait registers | `mountedModeRef` pre-check resolves immediately, so the signal cannot be missed |
| Stale signal from the outgoing pane | Deferred is keyed by mode; non-matching signals ignored |
| `AuthCard` unmounts mid-transition | Cleanup resolves any pending deferred |
| Strict Mode double-mount | `notifyPaneReady` is idempotent |
| Animation throws / interrupted | `try/finally` always clears the lock |

The watchdog is a **failsafe, not a sequencing mechanism** — it never fires on the happy path and never gates the reveal on elapsed time. No `setTimeout` paces any step.

## Notable implementation choices

- **`clip-path`, not skew + counter-skew.** The element stays an ordinary rectangle, so its translate is a single clean motion value, and nothing inside needs a counter-transform (clip-path clips paint without skewing internal layout). Edge glow uses `drop-shadow`, which follows the clipped silhouette — `box-shadow` would trace the unclipped rectangle.
- **`soft-light`, not `overlay`, for the artwork tint.** The gradient lightest stop is near-white and `overlay` blew the photograph out to flat colour entirely. Also added `isolation: isolate` so the blend does not composite against the blade opaque gradient underneath.
- **Artwork sits inside the counter-translated wrapper**, sized to the visible slice. Stretched across the full 118% blade it was upscaled ~2x from a 688px-wide portrait source and read as flat colour.
- **`.btn-primary` stays a flat fill.** It is used site-wide, so the gradient CTA is a new sibling `.btn-primary-gradient` rather than a change to the shared button.
- **`auth-royal` (`#3B4A78`) is deliberately not folded into the existing `royal` scale** (`#4361ee` family), which is a different blue reserved site-wide in CLAUDE.md.

Animation is Framer Motion throughout, using the named easings from `src/lib/motion.ts` — no new dependency, no inlined bezier arrays.

## Also in scope

`/forgot-password` and `/verify-email` restyled to the same palette (no blade — nothing to switch between). New tokens: `dusty-pink`, `off-white`, `soft-tan`, `dark-emerald`, `auth-royal`, plus the rose-gold gradient/sheen/glow utilities.

Password-strength colours keep the semantic weak→strong progression (it is a usability signal, same reasoning as red field errors) but are re-tinted from brand tokens, with "Strong" landing on an emerald tint.

**No auth logic changed** — `AuthContext`, `redirect.ts`, `constants.ts` and `proxy.ts` are untouched. `LoginContent`/`SignupContent` keep 100% of their business logic; only their `return` delegates to the new presentational panes.

## Verification

`npm run build` and lint pass clean.

**Desktop:** direct loads park correctly with no flash; both transition directions sweep → swap-under-blade → reveal; three rapid clicks produce exactly one transition; `?from=/checkout` survives a `/login → /signup → /login` bounce.

**Responsive**, measured at every width:

| Width | Result |
| --- | --- |
| 320, 390 | Band + form, nothing clipped, card grows (signup card 996px fully contains its form) |
| 768 | Mobile band, Y axis |
| **1023** | Fully mobile — Y axis, band clip, in-flow slot |
| **1024** | Fully desktop — identical to 1025, so the media query and the `lg:` classes agree *at* the boundary |
| **1025** | Fully desktop |
| 1440 | Unchanged from the desktop commit (`tx: -835`) |

Also passed: mobile dip/swap/retract landing at the re-measured rest for the taller signup card, rotation flipping the axis and zeroing the stale one, and a rapid triple-tap still yielding exactly one transition.

### Still worth a manual pass on the deploy preview

- **Real device.** This Chrome window clamps at 1280px wide, so narrow widths were tested in same-origin iframes — media queries inside an iframe respond to the iframe own width, so it is faithful, but it is not a real phone.
- **`prefers-reduced-motion`** — short-circuits to an instant `router.replace` on both axes, but the media query could not be toggled with the tooling available.
- **The watchdog failsafe** — only triggers on a route that never mounts, so it was not exercised.

One non-bug worth knowing: Framer Motion is rAF-driven, so a transition pauses if the tab is backgrounded mid-sweep and resumes on return. No stuck state.

https://claude.ai/code/session_01CrFFNGs8cEuaQfipRBoEYE


Oreginal:
## Code review fixes — `333fba4`

Ran a review over this branch. Seven findings, all addressed. Five I reproduced first so the fix could be measured rather than assumed.

| # | Issue | Sev | Status |
|---|---|---|---|
| 1 | Desktop signup form clipped once validation errors show | HIGH | Fixed |
| 2 | Blade snaps on the first frame of a transition | MED | Fixed |
| 3 | Blade paints untransformed before hydration | MED | Fixed |
| 4 | Mobile hydrates on the desktop axis | MED | Fixed |
| 5 | Toast lost its rounded corners | MED | Fixed |
| 6 | "Full cover" did not actually cover | LOW | Fixed |
| 7 | Mode toggles lost anchor semantics | LOW | Fixed |

---

### 1. Desktop signup form was clipped (HIGH)

**Issue.** The form slot was `lg:absolute lg:inset-0` inside an `overflow-hidden` card with a fixed `lg:min-h-[720px]`. An absolutely-positioned child contributes no height, so the card could never grow. Baseline signup fit with ~30px to spare — but the form *grows on interaction*:

| State | Form height | Card | Result |
|---|---|---|---|
| Baseline | 688px | 720px | fits (16px margin) |
| + password strength meter | 700px | 720px | fits (10px margin) |
| + three inline validation errors | **766px** | 720px | **clips 46px, both ends** |

With `lg:items-center` it cut ~23px off each end — taking the "Create your account" eyebrow and the top of the heading off the top, and the **"Already a member? Sign in" toggle** off the bottom. That is the only in-card route back to sign-in, and it disappeared precisely when the user had made a mistake. Reproduced at 1024, 1200, 1440 and 1600.

Notably this is the same class of bug the mobile commit fixed by moving the form into normal flow — desktop had been left behind.

**Fix.** The slot is now in normal flow at every breakpoint, so the card always grows with its content. Desktop keeps the mirrored composition by padding the side the blade rests on (`lg:pl-[54%]` / `lg:pr-[54%]`) rather than absolute positioning, and the 720px floor moved from the card onto the slot so it can both hold the floor and grow past it.

**Verified.** Error state now measures `cardH: 894`, `clippedTop: false`, `clippedBottom: false` at 1024/1280/1440, with the toggle sitting inside the card bounds.

---

### 2. Blade snapped on the first frame (MEDIUM)

**Issue.** `park()` wrote `element.style.transform` directly. But framer-motion caches a motion value per element on first `animate()` and from then on interpolates **from that cache, not from the DOM** — so every later `park()` write was invisible to it.

Reproduced: after a completed transition (framer cached `y=-796`), writing `translateY(-100px)` directly and triggering the next transition gave an early frame of **`-269`** — consistent with easing from the stale `-796`, not from `-100`. In practice that is a visible jump after any ResizeObserver re-park or a rotation.

**Fix.** Phase 1 now passes an explicit `[from, to]` keyframe pair, with `from` read off the DOM. Worth noting for anyone hitting this: seeding with a zero-duration `animate()` **does not work** — I tried it, and even awaiting `.finished` the stale cache still won. The explicit first keyframe is what actually forces the start point.

**Verified.** Sampled the curve through a transition after perturbing the DOM to `-100`:

```
before:  -100 → -598 → -165 → -32 → 0     ← jumps away, then eases
after:   -100 →  -33 →   -9 →  -5 → 0     ← eases directly
```

---

### 3 & 4. Pre-hydration paint and hydration axis (MEDIUM)

**Issue.** Two symptoms, one root cause — the resting position was applied from JS.

`useLayoutEffect` only guarantees "before paint" for its own commit; the streamed SSR HTML paints well before hydration. Until then the blade sat at `translate(0,0)`: on desktop covering the sign-in form, on mobile covering the whole card. I actually hit this during development and misattributed it to an unsettled render.

Separately, `useIsDesktop`'s `getServerSnapshot()` returned `true`, and React defers the store-consistency correction to a passive effect — so on a phone the hydration commit parked with `translateX(-59%)` on an element laid out as a full-width top band.

**Fix.** Resting position now lives in `.auth-blade` in `globals.css`, keyed off a `data-mode` attribute and the `lg` media query. The mobile rest is expressible exactly as `translateY(calc(-100% + 240px))`, because the blade's height is the card's plus `SLANT`.

This has a second benefit: rest now **re-resolves on its own** whenever the card changes height, since the percentage is relative to the blade's own box. That removed the park-every-render layout effect *and* the `ResizeObserver` entirely — net less code. The transition writes inline transforms while it runs and clears them when it settles, handing ownership back to CSS.

**Verified.** The server-rendered HTML carries `class="… auth-blade …"` and `data-mode="signin"` with **zero inline transforms**, so it is correct before any JS runs.

---

### 5. Toast lost its rounded corners (MEDIUM)

**Issue.** The gradient left border used `border-image`, which is not clipped by `border-radius` and fills all four edges. So the toast rendered as a hard rectangle with the gradient tinting the top, right and bottom hairlines too — on every failed sign-in or sign-up.

**Fix.** The stripe is painted as a background layer (`background-image` + `background-size: 3px 100%`) over the background colour. Backgrounds *are* clipped by the radius, so `rounded-2xl` survives and the gradient stays on the left edge.

---

### 6. "Full cover" did not cover (LOW)

**Issue.** At 118% blade width the clip geometry made covering both corners simultaneously impossible: covering the top-left needs `t ≤ -11%`, covering the bottom-right needs `t ≥ -13.2%`, and those only overlap once the blade is ≥128% of card width. At the shipped `cover: -8%` the card showed through a 3.5%-wide triangle at the top-left and 4.4% at the bottom-right *during the swap* — weakening the "swap while fully hidden" guarantee the whole transition is built around.

**Fix.** Blade widened to **132%**, with `BLADE_X`/`CONTENT_X` recomputed, and the image slice adjusted `41% → 36%` since it is a fraction of the blade (41% of 132% would have been 54% of the card instead of 48%).

**Verified.** At the exact cover frame on a 1200px card: covers **15.8px past the left edge** and **19.7px past the right**.

---

### 7. Mode toggles lost anchor semantics (LOW)

**Issue.** The login↔signup cross-links had become `<button>`, so middle-click, open-in-new-tab and no-JS navigation between the two pages were gone.

**Fix.** Back to `<Link href>` with `preventDefault()` + `requestSwitch` in `onClick`, passing modified clicks (cmd/ctrl/shift/middle) straight through to the browser. The `router.replace` history behaviour is unchanged — that was an intentional decision to avoid filling history with toggle spam, and only the anchor loss was a regression.

---

### Re-verification

Full matrix re-run on both routes at **320, 390, 768, 1023, 1024, 1025, 1440** — axis, clip shape, in-flow slot, card height and rest source all correct, crossover still clean (1023 fully mobile, 1024 matching 1025 fully desktop), nothing clipped, and every resting position coming from CSS rather than an inline style.

`npm run build` and lint pass clean.

Still carried forward as unverified: real-device testing (this Chrome clamps at 1280px wide, so narrow widths were tested in same-origin iframes), `prefers-reduced-motion`, and the watchdog failsafe.

Oreginal:
## Second review pass — `0bc5d50`

Re-ran the review against `333fba4` to confirm the first round of fixes landed.

**All seven earlier findings are closed, and this pass found no high or medium correctness bugs.** The reviewer independently re-derived the parts most likely to be wrong and confirmed them:

- Desktop `BLADE_X`/`CONTENT_X` exactly mirror the `@media (min-width:1024px)` rules, and `mobileTargets()` is the precise px equivalent of the CSS `translateY(calc(-100% + 240px))`.
- Cover genuinely covers on both axes at 132% (re-derived corners: −1.3% to +101.6% of card width).
- Reading the start point off the DOM makes the animation self-heal across a breakpoint crossing, and `releaseToCss` correctly hands rest back.
- The ready handshake has no ordering race on the happy path.

Six new low findings, all addressed below.

---

### Stale `mountedModeRef` could defeat the ready handshake

The one real bug. `mountedModeRef` was written on every pane mount but **never cleared**. If the watchdog ever won the race, `mode` advanced to the new route while `mountedModeRef` still named the *old* pane. Switching back then hit the `mountedModeRef.current === next` fast path in `waitForPaneReady`, resolved instantly, and uncovered before the incoming pane had mounted — exactly the failure the handshake exists to prevent, reintroduced on the degraded path.

Only reachable after an already-failed navigation, but the fix is small: the watchdog now clears both refs before resolving.

### Back-to-shop link changed colour mid-cover

The link paints at `z-30`, above the `z-20` blade, but keyed its colour to `mode` — which flips during phase 2, while the blade is at full cover. Going signin→signup it turned dark ink on top of the rose-gold panel for the entire 550 ms reveal; going the other way it stayed dark ink while the blade swept underneath it through most of phase 1. The ink variant is now suppressed while `isTransitioning`, so the colour matches whatever is actually beneath it.

### Small link text was under WCAG AA

`text-dusty-pink` on the off-white card measures **3.36:1**, below the 4.5:1 floor — on the only navigational links out of each form ("Forgot password?", the mode toggle, Terms/Privacy), plus the form subtitle and `FloatingInput`'s floated label.

Those move to `rose-dark`:

| | Ratio | AA 4.5:1 |
|---|---|---|
| dusty-pink `#b47792` on off-white | 3.36:1 | ✗ |
| **rose-dark `#8b5e52` on off-white** | **5.28:1** | ✓ |
| **rose-dark on white** (input label) | **5.50:1** | ✓ |

Dusty pink stays on hover states, icons, checkbox accents and the focus glow — non-text, and how the brief described its role. Worth noting the branch was already moving toward compliance here, not past it: the rose-gold this replaced was 2.3:1.

### Comment rot

Lower stakes individually, but this diff duplicates `BAND_H`, `SLANT` and the blade percentages across three files and keeps them in sync **only** through these notes — so stale ones are more load-bearing than usual. Fixed: a reference to `.auth-blade-clip` (renamed to `.auth-blade`), "the full 118% blade" (now 132%), and two contradictory adjacent comments about the artwork tint — one explaining `overlay`, the next explaining why it deliberately isn't `overlay`.

`useIsDesktop`'s doc block also still described the layout-effect parking that CSS-owned rest replaced, and claimed a pre-paint reconciliation `useSyncExternalStore` doesn't actually provide. Rewritten to say why the hook *is* safe — nothing renders from it — rather than the opposite, since that's the reading most likely to cause a wrong change later.

### Documented, not changed

The toggle's handler calls `router.replace` while the `<Link href>` it overrides pushes, so a JS click and a middle-click diverge, and Back from signup skips the sign-in form rather than returning to it (`/checkout → /login?from=/checkout → "Create an account" → Back` leaves the auth flow).

That trade is deliberate — `replace` keeps repeated toggling out of the history stack — so it's now recorded in a comment rather than left looking like an oversight. The anchor is kept because it's what makes the toggle work without JS and in a new tab.

---

### Verification

Round trips re-checked on both axes:

```
@390   rest y=-659 (CSS) → signup y=-796 (CSS) → back y=-659   toggle=<A>
@1280  rest x=-919 (CSS) → signup x=+539 (CSS) → back x=-919   toggle=<A>
```

Every resting position sourced from CSS with no inline transform left behind, and the toggle renders as a real anchor. `npm run build` and lint pass clean.

Unchanged from before: real-device testing, `prefers-reduced-motion` and the watchdog failsafe remain unverified here and are worth a look on the deploy preview.

Issue #63, opened 2026-08-10
Wishlist and cart disagree on item id, producing duplicate bag rows


Oreginal:
Found while taking over #48 (see #62). Deliberately left out of that PR — the fix touches the cart and
wishlist contexts, which #62 otherwise leaves alone.

## Problem

`toCartItem()` in `src/lib/shopify.ts` builds a variant-keyed id — `product:1#4567` — whenever the
product has variants. The wishlist path does not: `WishlistHeart.handleWish` (in
`ShopifyProductCard.tsx`) and `handleWish` in `ProductDetailContent.tsx` both store the bare
`product.id` (`product:1`), and `WishlistSidebar.handleMoveToBag` then calls `addItem(item)` with that
bare id.

So for the same product:

- quick-add from a card → cart line `product:1#4567`
- move-to-bag from the wishlist → cart line `product:1`

Two rows in the bag for one product, two independent quantities, and the "in cart" state disagrees
between surfaces.

## Fix options

1. Have the wishlist store the variant-keyed id too (consistent, but a wishlist entry is arguably a
   product not a variant).
2. Have `WishlistSidebar.handleMoveToBag` route through `toCartItem()` so the id is built the same way
   the cart always builds it. Probably the smaller and more correct change.

Either way `CLAUDE.md`'s cart-id convention should be updated to say which surface owns the variant
suffix.

## Note

Pre-existing in the design that came over from #48 — not introduced by #62.

PR #74, opened 2026-08-15
Kat/feat/4/shopping cart func #4

KatlehoMadaba:


.github/workflows/restrict-pr-source.yml
```diff
@@ -0,0 +1,20 @@
...
+      - master
+
+jobs:
+  check-source-branch:
+    name: Only main may PR into master
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify head branch
+        run: |
+          if [ "${{ github.head_ref }}" != "main" ]; then
```
Oreginal, line 16: **HIGH — shell injection that also defeats the guard entirely.**

`${{ github.head_ref }}` is interpolated straight into the `run:` script here (and again on line 17). GitHub expressions are substituted as *text* before the shell ever sees them, and branch names permit `"`, `;`, `$`, backtick and `|`.

A PR from a branch named `x"; exit 0; #` expands line 16 to:

```sh
if [ "x"; exit 0; #" != "main" ]; then
```

`[` fails as a syntax error, `exit 0` runs, the step goes green — and the "only main may PR into master" rule is bypassed by anyone who can push a branch. Since this workflow is the thing protecting production, that's a full bypass of the branch policy, and arbitrary commands run on the runner as a bonus.

Fix — pass through `env:` so the value reaches the shell as data, never as script text:

```yaml
      - name: Verify head branch
        env:
          HEAD_REF: ${{ github.head_ref }}
        run: |
          if [ "$HEAD_REF" != "main" ]; then
            echo "PRs into master are only allowed from 'main'. This PR is from '$HEAD_REF'."
            exit 1
          fi
          echo "OK: PR is from main."
```

.github/workflows/restrict-pr-source.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+    branches:
+      - master
+
+jobs:
+  check-source-branch:
+    name: Only main may PR into master
```
Oreginal, line 11: **MEDIUM — the check reads the branch *name*, not the repository.**

`github.head_ref` is only the branch name, so it can't tell *your* `main` from a fork's `main`. Anyone can fork, push a branch literally named `main`, and open a PR straight into production `master` — the job passes.

Worth pairing the name check with a repo check:

```yaml
        if: github.event.pull_request.head.repo.full_name != github.repository
```

…or asserting `head.repo.full_name == github.repository` inside the step alongside the branch comparison.

src/app/api/revalidate/route.ts
```diff
@@ -1,11 +1,47 @@
...
+  const expected = Buffer.from(digest);
+  const received = Buffer.from(signature);
+  if (
+    expected.length !== received.length ||
+    !crypto.timingSafeEqual(expected, received)
+  ) {
+    return NextResponse.json({ message: "Invalid signature" }, { status: 401 });
   }
 
   revalidateTag("products", "max");
```
Oreginal, line 47: **HIGH — `"max"` never actually expires the tag, so the webhook doesn't bust the cache.**

This line is carried over unchanged, but it undercuts the whole point of the PR. The `"max"` profile resolves to `expire: 31_536_000` — it schedules expiry a *year* out and only marks the tag stale.

Concrete failure: a merchant changes a price → Shopify fires the webhook → this route now correctly authenticates it and returns `{revalidated: true}` → **the next shopper is still served the old price.** The endpoint reports success while doing nothing useful.

```ts
revalidateTag("products", { expire: 0 });
```

This is called out explicitly in `CLAUDE.md`, and it's what `main` already does.

src/app/api/revalidate/route.ts
```diff
@@ -1,11 +1,47 @@
...
+ * Authenticated by Shopify's own HMAC signature rather than a URL secret: every
+ * webhook is signed with `X-Shopify-Hmac-Sha256` (base64 HMAC-SHA256 of the raw
+ * request body, keyed by the webhook signing secret). Verifying that keeps the
+ * secret out of the query string (which leaks into access logs and Referer) and
+ * proves the body is untampered. `SHOPIFY_REVALIDATE_SECRET` must hold the same
+ * signing secret Shopify uses. Unset → fail closed (every request 401s; only the
+ * hourly fallback in getFeaturedProducts refreshes).
+ */
+export async function POST(req: Request) {
+  const secret = process.env.SHOPIFY_REVALIDATE_SECRET;
```
Oreginal, line 20: **MEDIUM — the meaning of `SHOPIFY_REVALIDATE_SECRET` changed, but the name didn't.**

The old contract was "a shared secret *you* invent, passed as `?secret=`". The new contract is "the webhook signing secret **Shopify generates**, shown under Settings → Notifications → Webhooks." Same variable name, incompatible value.

So any environment that already has `SHOPIFY_REVALIDATE_SECRET` set to a self-generated string keeps deploying fine and then **401s every webhook forever**, silently — prices just quietly stop updating, and nothing surfaces the mismatch. This is exactly the kind of break that gets found weeks later.

Renaming to `SHOPIFY_WEBHOOK_SECRET` (as `main` does) forces whoever deploys to set the correct value rather than inheriting a stale one. The `.env.example` comment should spell out that the value comes from Shopify, not from you.

src/app/api/revalidate/route.ts
```diff
@@ -1,11 +1,47 @@
...
+  if (!secret) {
+    return NextResponse.json({ message: "Not configured" }, { status: 401 });
```
Oreginal, line 22: **LOW — an unset secret is indistinguishable from an attack, and tells Shopify to give up.**

Missing config returns the same bare 401 as a forged request and logs nothing, so in Shopify's delivery log a misconfigured deploy looks identical to someone probing the endpoint. 401 also reads as "your request is wrong, don't bother retrying" rather than "this end is broken, try again later."

```ts
if (!secret) {
  console.error("SHOPIFY_WEBHOOK_SECRET is not set — rejecting webhook");
  return NextResponse.json({ message: "Webhook not configured" }, { status: 503 });
}
```

Diagnosable, and Shopify keeps retrying so deliveries aren't lost once it's fixed.

src/app/store/[slug]/ProductDetailContent.tsx
```diff
@@ -52,6 +52,10 @@ export default function ProductDetailContent({
   const price = variant?.price ?? product.price;
   // Shopify's compareAtPrice is a product-level range, so it only lines up with
   // the variant the range was derived from — show it on the cheapest variant.
+  // Known limitation: if that cheapest variant is sold out, defaultVariant()
```
Comment (Oreginal) [inline src/app/store/[slug]/ProductDetailContent.tsx:55]: **LOW — the follow-up points at the issue this PR closes.**

The limitation documented here is real and worth writing down: when the cheapest variant is sold out, `defaultVariant()` picks a pricier in-stock one, `price !== product.price`, and the strikethrough silently disappears from a genuinely discounted product.

But "follow-up: track in #4" points at the issue this PR itself closes — the moment this merges, #4 closes and the tracking is lost. Worth opening a small issue for the per-variant `compareAt` work and referencing that number instead.

Oreginal (commented):
## Code review — PR #74

Reviewed the 6-file diff against `main` (+74/−24). Five findings below, two of them high severity.

**Before the line notes — a merge-base problem that affects most of this PR.** This branch is based on a stale merge base, and `main` already contains a superseding implementation of `src/app/api/revalidate/route.ts`: the hardened version with `revalidateTag("products", { expire: 0 })`, `SHOPIFY_WEBHOOK_SECRET`, and a 503 on missing config. That file will conflict on merge, and resolving it the wrong way would regress `main`. Rebasing onto `main` will drop most of the route work here as already-done, and resolves findings 2–4 outright.

The HMAC approach in this PR is the right call — keeping the secret out of the query string (where it leaks into access logs and `Referer`) and signing the raw bytes are both correct. It's just that `main` already landed the same idea with the rough edges filed off.

_Reviewed with Claude Code._

Oreginal:
## Correction: I overstated finding 1

In [my review above](https://github.com/KatOre-Solutions/HeyBeautiful/pull/74#pullrequestreview-4972993888) I marked the `${{ github.head_ref }}` interpolation in `restrict-pr-source.yml` as **HIGH — shell injection that also defeats the guard entirely**, and claimed a branch named `x"; exit 0; #` would make the step go green and bypass the branch policy.

**That claim is wrong.** I tested it against the real script and it fails **closed**, not open.

Bash parses the whole `if…fi` compound before executing any of it, so a malformed branch name is a syntax error — nothing executes, the step exits non-zero, and the check *fails*. My asserted payload produces exit 2, not 0:

| Branch name | Exit | Result |
| --- | --- | --- |
| `x"; exit 0; #` | 2 | fails closed |
| `x";exit${IFS}0;#` | 2 | fails closed |
| `x";exit${IFS}0;if${IFS}true;then${IFS}#` | 2 | fails closed |
| `feature-branch` | 1 | correctly blocked |
| `main` | 0 | correctly allowed |

Git's ref-name rules help too — no spaces, no `[`, no `:` — which removes the usual ways of reopening a `test` or supplying a no-op. I could not construct a payload that both parses and reaches a green `exit 0`.

**Revised severity: LOW / hardening.** Interpolating untrusted input into `run:` is still the `actions/command-injection` anti-pattern that CodeQL flags and GitHub's hardening guidance warns against, and it should still be passed via `env:`. But it is not an open door to production, and I shouldn't have described it as one.

**Finding 2 is unaffected and is the one that matters.** `github.head_ref` is only a branch *name*, so a fork branch called `main` satisfies the check and can PR straight into `master`. That needs no exploit construction — it just works.

Sorry for the noise. Both fixes are now in #79 against `main`, so this file can simply be dropped from this PR if that's easier — note #79 and this PR add the same file, so whichever lands second will conflict and the hardened version should win.

The other five findings stand as written.

Oreginal:
## Closing this branch — the work it asked for has all shipped

Reviewed this against current `main` before deciding. Summary up front: **all three concerns raised
here were correct and all three are now on `main`**, implemented independently rather than through
this branch. The branch itself is 94 commits behind (forked at `898d1e9`) and can no longer merge.

### The three concerns, and where each landed

| Concern raised here | Where it shipped |
| --- | --- |
| The `?secret=` query-string scheme was insecure | **`32375fc` (#4)** — HMAC verification of `X-Shopify-Hmac-Sha256` with `timingSafeEqual` |
| `master` needed protecting from arbitrary source branches | **`f6dd4da` + `7d1fd05` (#78)** — `restrict-pr-source.yml`, now a required check |
| Prices were formatted inconsistently | **`formatPrice`** is now the only formatter; the rule is written into `CLAUDE.md` |

The diagnosis on all three was right. That is the part that mattered.

### Why the branch can't be merged now

Per-file, against `main` as of today:

| File | Status |
| --- | --- |
| `.env.example` | Superseded — the variable was renamed to `SHOPIFY_WEBHOOK_SECRET`, and `.env.example` now tells you to delete the old one |
| `.github/workflows/restrict-pr-source.yml` | Superseded — main's version adds `permissions: {}`, passes untrusted values as quoted env vars, and adds a same-repo check that closed a real fork bypass (#78) |
| `src/app/api/revalidate/route.ts` | Superseded — main fails closed with 503 and uses `revalidateTag("products", { expire: 0 })`. This branch's `"max"` only marks the tag stale, so the next shopper still gets the old price |
| `src/app/store/[slug]/ProductDetailContent.tsx` | Would **revert** two fixes — see below |
| `src/components/ShopifyProductCard.tsx` | Already on main |
| `src/components/WishlistSidebar.tsx` | Already on main |

The `ProductDetailContent.tsx` one is the reason a rebase isn't the answer either. Merging as-is
would delete `cartLine: toCartItem(product, variant)` — the #63 fix that stops moving an item from
the wishlist opening a *second* bag row. Since #92, wishlist entries without `cartLine` are dropped
on restore, so removing it is worse now than when this branch was written. It would also revert
main's per-variant `originalPrice` fix back to a comment describing that bug as a known limitation —
main has since actually fixed it.

### It would also fail CI

Two independent failures on the required `Lint · Typecheck · Test · Build` check:

1. **Typecheck** — two files import `defaultVariant`, `isSoldOut`, `toCartItem` and `ShopifyProduct`
   from `@/lib/shopify`. Those moved to `@/lib/product`; `shopify.ts` exports none of them today.
2. **Lint** — `scripts/check-images.mjs` asserts the literal strings `priority={priority}` and
   `priority={index < 4}`. This branch's `ShopifyProductCard` predates the `variant`/`priority`
   props and takes only `{ product }`.

### Still in the codebase

`5429f0e` — *"feat: add Shopify product data layer with on-demand revalidation (#2)"* — is on `main`
and is the foundation `src/lib/shopify.ts` still rests on today, including the catalogue reads the
checkout handoff was built against.

### Housekeeping

Closing rather than rebasing, since there is nothing left to carry over. **The branch is being left
in place**, not deleted. Issue #4 stays open — its last unticked box is the checkout handoff, which
is PR #94, not anything from here.

Issue #59, opened 2026-07-20
Add GitHub Actions CI pipeline (lint, typecheck, test, build)


Oreginal:
## Problem

There is **no CI** in this repo — no `.github/workflows`, no pre-commit hooks. Quality checks (`lint`, `typecheck`, `build`) only run when someone remembers to type them locally. This was explicitly deferred as out-of-scope in #52:

> **Out of scope:** Adding CI. Worth its own issue — without it, `lint` still only runs when someone remembers to type it.

This is that issue.

## Goal

A production-ready GitHub Actions workflow at `.github/workflows/ci.yml` that gates the long-lived branches, matching the repo's promotion flow:

```
feature/* ──PR──▶ main ──PR──▶ master (production; Netlify deploys)
```

## Scope

- **Triggers:** push to `main` / `master`; PRs targeting `main` / `master`.
- **Node:** pinned via a new `.nvmrc` (there is none today) + `engines`.
- **Package manager:** npm (`npm ci`), with dependency + Next.js build caching.
- **Steps, stop-on-first-failure, cheapest first:**
  1. `lint` (`eslint . --max-warnings 0` + image guard)
  2. `typecheck` (new `tsc --noEmit` script — TS is used but had no type gate)
  3. `test` (`--if-present`; no tests exist yet, wired for when they do)
  4. `build` (`next build` — validated to succeed with no secrets)
- **Dependabot** (`.github/dependabot.yml`): weekly npm + github-actions updates.

## Dependency

⚠️ **Depends on #52 (PR #58).** CI runs `npm run lint`, which only works after the lint pipeline is repaired in #52. This must merge **after** #58, or the first run on `main` fails at the lint step.

## Acceptance criteria

- [ ] `.github/workflows/ci.yml` runs lint → typecheck → test → build, stopping on first failure
- [ ] Runs on push + PR to `main` and `master`
- [ ] Node version single-sourced (`.nvmrc`), deps + Next build cached
- [ ] `typecheck` script added (`tsc --noEmit`)
- [ ] Dependabot configured
- [ ] Branch protection updated: required status check `Lint · Typecheck · Test · Build` on `main` and `master` (after first run completes)

## Out of scope

Prettier/format checks, CodeQL security scanning, and code coverage — all recommended as follow-ups (coverage is premature until tests exist).

Issue #57, opened 2026-07-19
Enforce email_verified server-side on order/payment endpoints (not just the client checkout gate)


Oreginal:
## Context

PR #49 (closes #22) enforces email verification at `/checkout` **client-side** — the edge `proxy.ts` can't read Firebase's `emailVerified`, so the gate lives in `CheckoutContent`. That's correct as scoped, and it's the right UX.

But it is **UX, not a security boundary.** A client-side redirect is trivially bypassable: disable JS, call the order/payment API directly, or edit client state. Raised by @KatlehoMadaba in the #49 review.

The #49 description frames the checkout gate as establishing a *"reached checkout ⇒ verified email"* invariant that *"future order/payment logic will rely on."* **That reliance must be enforced server-side** — the client redirect must not be mistaken for the enforcement point.

## What to do (when the order/payment surface is built)

There's no order/payment surface in the repo yet, so nothing to change today. When it lands, verify `email_verified` at the trust boundary:

- [ ] On the order/payment **endpoint**: verify the Firebase **ID token** and reject unless `email_verified === true` (don't trust a client-sent flag).
- [ ] And/or in **Firestore security rules**: `request.auth.token.email_verified == true` on order/payment writes.
- [ ] Treat the `/checkout` client gate purely as UX — it must not be the only thing standing between an unverified user and a real order.

## Links

- Follows from #22 / PR #49 (the client-side gate).

Oreginal:
Minor follow-up parked here from the #49 review (unrelated to the server-side enforcement above, just tracking it so it isn't lost):

**`reloadUser()` error spam on a dead session.** In `VerifyEmailContent`'s poll loop, if the session is expired/revoked, `currentUser.reload()` rejects every tick and `console.error`s every ~3s until `onAuthStateChanged(null)` fires the mount guard to `/login`. Harmless but noisy. Fix when poll backoff is added: swallow or back off the rejection on the poll path. Not worth a standalone issue.

PR #76, opened 2026-08-15
fix: harden the authentication workflow

### LINKED ISSUE #75: Harden the authentication workflow (redirects, session, verification, errors)
## Summary

Harden the Firebase authentication workflow across redirect handling, session state, the
unverified-user journey, verification reliability, signup recovery, social credential
conflicts, password-reset privacy, error handling, and auth UX.

Delivered as ten reviewed phases. Existing behaviour was preserved throughout: Firebase Auth
stays the provider, the popup→redirect social fallback is unchanged, `/account` remains
reachable while unverified, and the checkout verification requirement is unchanged.

## Scope

1. **Redirect security** — one validated, non-throwing `resolveDestination`; `withFrom` as the
   security boundary; a single documented precedence + sessionStorage lifecycle.
2. **Session state** — presence cookie renamed and documented as a routing hint only; cookie
   lifetime matched to Firebase persistence; stale-hint clearing; redirect-loop breaker.
3. **Unverified journey** — one source of truth for which routes require verification.
4. **Verification reliability** — visibility-aware polling with a deadline; 60s resend cooldown.
5. **Signup recovery** — critical vs recoverable failure separated; duplicate-email recovery
   path; email normalisation; 8-character password policy.
6. **Social conflicts** — `account-exists-with-different-credential` handled with a route back
   in; Apple's first-authorization-only display name captured; redirect errors surfaced.
7. **Password-reset privacy** — account-enumeration oracles closed on both the reset *and*
   sign-in forms.
8. **Error handling** — all Firebase error interpretation centralised in `src/lib/auth-errors.ts`.
9. **Auth UX** — password strength aligned to the enforced policy; session-expiry messaging;
   fuller verification reminder.
10. **Account management foundation** — assessed and documented only; see
    `docs/auth-account-management.md`.

## Known limitations (deliberately out of scope)

- No server-side enforcement of `emailVerified` — tracked separately in #57. There are no
  server-side user operations yet (checkout is a placeholder), so client-side gating is
  currently the whole model. This becomes load-bearing when payments land.
- Provider linking/unlinking is designed and documented, not built.
- A timing side channel remains on password reset; closing it needs a server-side proxy.

Oreginal:
Closes #75

Ten-phase hardening pass over the Firebase auth flow, reviewed phase by phase.

**Preserved deliberately:** Firebase Auth stays the provider, the popup → redirect social
fallback is unchanged, `/account` remains reachable while unverified, and the checkout
verification requirement is untouched.

## The four bugs that were live

1. **Open redirect.** The guard only blocked the literal prefixes `//` and `/\`, so
   `/%09//evil.com` and `/%5C%5Cevil.com` passed — browsers normalise tabs and backslashes
   *after* our check ran. Separately, the `sessionStorage` fallback reached `router.push`
   with no validation at all.
2. **Signup reported failure for accounts that existed.** `updateProfile` sat unguarded
   between account creation and the guarded verification send. If it threw, the user was
   told signup failed and retried straight into `auth/email-already-in-use`.
3. **Account enumeration on two forms.** "No account found with this email" and "Incorrect
   password" let anyone test an address for membership — on the sign-in form as well as
   password reset.
4. **Raw internal errors rendered to users.** `getAuthErrorMessage` fell back to
   `err.message`, so a stray `TypeError` from our own code would surface verbatim.

## Structural changes

| Area | Change |
| --- | --- |
| Redirects | `src/lib/redirect.ts` rewritten as a total, non-throwing boundary. `withFrom` validates its own input, so caller discipline can't leak an unsafe destination. |
| Precedence | One documented rule: a present-but-invalid `from` resolves to `/account` and **never** falls through to stored state. |
| Session hint | `AUTH_COOKIE` → `SESSION_HINT_COOKIE`, documented as routing-only. Lifetime now matches Firebase persistence, so "remember me" off no longer leaves a 30-day stale hint. |
| Redirect loop | The proxy no longer bounces auth pages carrying `?from=` — a hint the client can't clear used to ping the user between `/account` and `/login` forever. |
| Sign-out | Clears cart and wishlist storage; they previously survived on shared devices. Session *expiry* deliberately does not. |
| Errors | All Firebase error interpretation in `src/lib/auth-errors.ts`, duck-typed rather than `instanceof` so it survives duplicate SDK copies in a bundle. |

## Verification

- **275 assertions** run against the real modules via Node type-stripping — hostile redirect
  inputs, the enumeration-safe message trio reducing to a set of size 1, provider name shapes
  for Google *and* Apple, password strength, session-expiry URL building.
- Browser-verified without credentials: proxy redirect fidelity, stale-hint clearing with no
  loop, malformed `%zz` / `%E0%A4%A` on all four auth routes with zero console errors,
  invalid `from` not consuming stored state, client-side validation making **zero**
  `identitytoolkit` requests, neutral reset confirmation for an unregistered address, and the
  password meter's computed colours.
- `npm run build` passes, TypeScript clean.

**`npm run lint` still fails on two pre-existing `react-hooks/rules-of-hooks` errors in
`src/lib/shopify.ts`** (`usePlaceholderCatalogue`). Confirmed present on `main` by stashing
this branch's changes — not introduced here, but it means lint is currently useless as a gate
(relates to #52 / #59).

## Not verified — needs real accounts

Anything requiring a signed-in, unverified, or social account: the verification polling and
resend, signup partial-failure recovery, credential conflicts, Apple's name capture, and the
`/account` reminder card. The logic behind these is covered by the assertions above; the
Firebase integration around them is not. A manual checklist is in the phase reports.

## Scope notes

- **Phase 10 is assessment only** — `docs/auth-account-management.md` documents provider
  linking/unlinking, required Firebase APIs, recent-login handling and edge cases. Nothing
  was built, because there is no Account Settings screen to build it into (the `/account`
  cards are still stubs).
- Three error codes for provider linking are mapped with no callers yet, flagged in-file.
  Easy to drop if you'd rather keep the map to what can currently fire.
- **Server-side `emailVerified` enforcement is still open as #57.** There are no server-side
  user operations yet — checkout is a placeholder — so client-side gating is currently the
  entire model. This becomes load-bearing the moment payments land.

https://claude.ai/code/session_01FqMNJw3rzcQr6FJFa4Udmv

### DIFF
--- a/src/app/(auth)/(card)/login/LoginContent.tsx
+++ b/src/app/(auth)/(card)/login/LoginContent.tsx
@@ -111,6 +111,11 @@ function LoginForm() {
     e.preventDefault();
     if (loading) return;
     setError("");
+    // A conflict raised by a social attempt is about a different sign-in method — leaving
+    // it up would pin that panel over an unrelated email/password failure. Both sources
+    // have to go, since credentialConflict falls back to the redirect-path one.
+    setPopupConflict(null);
+    if (redirectError) clearRedirectError();
     setLoading(true);
     try {
       await signIn(email, password, remember);
--- a/src/app/(auth)/(card)/login/LoginContent.tsx
+++ b/src/app/(auth)/(card)/login/LoginContent.tsx
@@ -1,34 +1,60 @@
 "use client";
 
-import { Suspense, useEffect, useMemo, useState } from "react";
+import { Suspense, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useRouter, useSearchParams } from "next/navigation";
 import { useAuth } from "@/context/AuthContext";
 import SignInPane from "@/components/auth/SignInPane";
 import { useAuthTransition } from "@/components/auth/AuthTransitionContext";
-import AuthErrorToast, { getAuthErrorMessage } from "@/components/auth/AuthErrorToast";
-import { REDIRECT_KEY } from "@/lib/constants";
-import { resolveDestination } from "@/lib/redirect";
+import AuthErrorToast from "@/components/auth/AuthErrorToast";
+import {
+  getAuthErrorMessage,
+  getConflictEmail,
+  isCredentialConflictError,
+} from "@/lib/auth-errors";
+import {
+  DEFAULT_DESTINATION,
+  SESSION_EXPIRED_PARAM,
+  clearDestination,
+  postAuthDestination,
+  readDestination,
+  resolveDestination,
+  saveDestination,
+} from "@/lib/redirect";
 
 function LoginForm() {
   const router = useRouter();
   const params = useSearchParams();
   const { setLastLoginFrom } = useAuthTransition();
   const fromParam = params.get("from");
-  const destination = useMemo(() => {
-    if (fromParam) return resolveDestination(fromParam);
-    const saved =
-      typeof window !== "undefined"
-        ? sessionStorage.getItem(REDIRECT_KEY)
-        : null;
-    return saved ?? "/account";
-  }, [fromParam]);
-  const { signIn, signInWithGoogle, signInWithApple, user, loading: authLoading } = useAuth();
+  // A protected page found the session gone and sent them here. Say so, rather than leaving
+  // them to wonder why they're suddenly looking at a sign-in form.
+  const sessionExpired = params.get(SESSION_EXPIRED_PARAM) === "expired";
+  // Precedence (documented in src/lib/redirect.ts): a `from` on the URL always decides,
+  // *including* when it is invalid — falling through to the stored value would let a
+  // hostile `?from=` interact with stale client state. Storage is consulted only when no
+  // `from` was supplied at all, which is the social-redirect return.
+  const destination = useMemo(
+    () =>
+      fromParam !== null
+        ? resolveDestination(fromParam)
+        : readDestination() ?? DEFAULT_DESTINATION,
+    [fromParam]
+  );
+  const {
+    signIn,
+    signInWithGoogle,
+    signInWithApple,
+    user,
+    loading: authLoading,
+    redirectError,
+    clearRedirectError,
+  } = useAuth();
 
-  // Stash it on the card so a /login → /signup → /login bounce keeps the
-  // destination; AuthCard outlives both routes, so the value survives.
+  // Stash the resolved destination on the card so a /login → /signup → /login bounce keeps
+  // it; AuthCard outlives both routes, so the value survives the swap.
   useEffect(() => {
-    if (fromParam) setLastLoginFrom(fromParam);
-  }, [fromParam, setLastLoginFrom]);
+    if (destination !== DEFAULT_DESTINATION) setLastLoginFrom(destination);
+  }, [destination, setLastLoginFrom]);
 
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
@@ -36,17 +62,50 @@ function LoginForm() {
   const [remember, setRemember] = useState(true);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState("");
+  // Conflicting address, or "" when Firebase didn't name one. null = no conflict.
+  const [popupConflict, setPopupConflict] = useState<string | null>(null);
 
+  // Turns a popup-path failure into UI. The redirect path is derived below instead.
+  const reportAuthError = useCallback((err: unknown) => {
+    if (isCredentialConflictError(err)) {
+      setPopupConflict(getConflictEmail(err) ?? "");
+      return;
+    }
+    const message = getAuthErrorMessage(err);
+    if (message) setError(message);
+  }, []);
+
+  // A redirect sign-in fails *before* this page reloads, so its error arrives through
+  // context rather than a catch. Derive the display from it rather than copying it into
+  // local state — the context already holds it, and mirroring would only add a render pass.
+  const redirectConflict = isCredentialConflictError(redirectError)
+    ? (getConflictEmail(redirectError) ?? "")
+    : null;
+  const credentialConflict = popupConflict ?? redirectConflict;
+  const errorMessage =
+    error ||
+    (redirectError && !isCredentialConflictError(redirectError)
+      ? getAuthErrorMessage(redirectError)
+      : "");
+
+  const dismissError = () => {
+    setError("");
+    if (redirectError) clearRedirectError();
+  };
+
+  // The single navigation path. Email sign-in, popup sign-in and the post-redirect return
+  // all land here once Firebase reports a user, so the destination is resolved once and the
+  // stored value consumed exactly once. Ref-guarded because Strict Mode double-invokes.
+  //
+  // postAuthDestination is what routes an unverified user heading for /checkout through
+  // /verify-email first, instead of letting checkout bounce them there.
+  const navigated = useRef(false);
   useEffect(() => {
-    if (authLoading || !user) return;
-    const dest =
-      (typeof window !== "undefined"
-        ? sessionStorage.getItem(REDIRECT_KEY)
-        : null) ?? destination;
-    sessionStorage.removeItem(REDIRECT_KEY);
-    router.push(dest);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [user, authLoading]);
+    if (authLoading || !user || navigated.current) return;
+    navigated.current = true;
+    clearDestination();
+    router.push(postAuthDestination(destination, user.emailVerified));
+  }, [user, authLoading, destination, router]);
 
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
@@ -55,7 +114,8 @@ function LoginForm() {
     setLoading(true);
     try {
       await signIn(email, password, remember);
-      router.push(destination);
+      // Navigation is the effect above's job — pushing here too raced it and could send
+      // the user somewhere different from the social paths.
     } catch (err) {
       setError(getAuthErrorMessage(err));
     } finally {
@@ -64,16 +124,20 @@ function LoginForm() {
   };
 
   const handleSocial = async (fn: () => Promise<void>) => {
-    sessionStorage.setItem(REDIRECT_KEY, destination);
+    // Only the social paths write storage: signInWithRedirect is a full-page navigation
+    // that destroys this component's state, so the destination needs somewhere to live.
+    saveDestination(destination);
     setError("");
+    setPopupConflict(null);
+    // A stale failure from a previous redirect shouldn't linger over a fresh attempt.
+    if (redirectError) clearRedirectError();
     try {
       await fn();
-      // useEffect([user, authLoading]) handles navigation and cleanup for both
-      // popup-success and post-redirect-return paths
+      // The effect above handles navigation and cleanup for both popup-success and
+      // post-redirect-return paths.
     } catch (err) {
-      sessionStorage.removeItem(REDIRECT_KEY);
-      const msg = getAuthErrorMessage(err);
-      if (msg) setError(msg);
+      clearDestination();
+      reportAuthError(err);
     }
   };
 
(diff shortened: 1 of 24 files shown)


src/app/account/AccountContent.tsx
```diff
@@ -38,19 +48,54 @@ const cards = [
...
 
-  // The proxy gates this route on a presence cookie only. If the Firebase session
-  // is actually gone (expired / stale cookie), bounce to login rather than render
-  // the dashboard to a logged-out visitor.
+  // The proxy gates this route on the presence hint alone, which proves nothing. Now that
+  // Firebase has spoken, enforce the real answer: no user means the hint was stale, expired
+  // or hand-set, so clear it and bounce — carrying the destination so signing back in
+  // returns here. Clearing before navigating is what stops the proxy seeing the dead hint
+  // on /login and bouncing them straight back.
   useEffect(() => {
```
Oreginal, line 96: **MEDIUM — an explicit sign-out trips the new "session expired" guard.**

`handleSignOut` awaits `signOut()`, which drives `onAuthStateChanged` → `setUser(null)`, and only *then* calls `router.push("/")`. This effect fires on that same `user === null` render, and nothing here distinguishes "the user deliberately signed out" from "the hint was lying".

So the guard's `router.replace(sessionExpiredLoginUrl("/account"))` races the handler's `router.push("/")` — and being issued from a passive effect *after* the push, it typically wins.

Scenario: a user clicks **Sign Out** on `/account` and lands on (or at minimum flashes) the sign-in form reading *"Your session has ended. Please sign in again — we'll take you straight back to where you were."* They did not expire; they left on purpose.

The old code had the same race, but it sent them to a plain `/login?from=/account`. The new `session=expired` copy upgrades a harmless race into an actively false statement, which is what makes this worth fixing now.

A `signingOut` ref — or simply setting `redirected.current = true` before `await signOut()` — settles it deterministically.

src/app/account/AccountContent.tsx
```diff
@@ -38,19 +48,54 @@ const cards = [
 
 export default function AccountContent() {
   const router = useRouter();
-  const { user, loading, signOut } = useAuth();
+  const { user, loading, signOut, sendVerification } = useAuth();
+  const { clearCart } = useCart();
+  const { clearWishlist } = useWishlist();
+  const cooldown = useCooldown();
+  const [resending, setResending] = useState(false);
+  const [resent, setResent] = useState(false);
```
Oreginal, line 59: **LOW — `resent` is never reset, so the resend button sticks permanently.**

`setResent(true)` on success has no counterpart anywhere in this file. The label chain is:

```
resending → resent → cooldown.active → default
```

so `resent` outranks the cooldown. After one resend the button reads **"Verification email sent ✓"** forever. Once the 60s cooldown lapses it becomes clickable again, but the label never changes — so a second resend gives the user *no feedback at all* and reads as broken.

`VerifyEmailContent` already solves this correctly with `RESENT_NOTICE_MS`:

```ts
useEffect(() => {
  if (!resent) return;
  const id = setTimeout(() => setResent(false), RESENT_NOTICE_MS);
  return () => clearTimeout(id);
}, [resent]);
```

That effect is simply missing here.

src/app/(auth)/(card)/signup/SignupContent.tsx
```diff
@@ -88,7 +196,14 @@ export default function SignupContent() {
...
+        onEmailChange={(value) => {
+          setEmail(value);
+          // Editing the address makes the "already registered" notice stale.
+          if (emailAlreadyInUse) setEmailAlreadyInUse(false);
+          if (fieldErrors.email) setFieldErrors((f) => ({ ...f, email: undefined }));
+        }}
+        emailAlreadyInUse={emailAlreadyInUse}
+        credentialConflict={credentialConflict}
         password={password}
         onPasswordChange={setPassword}
```
Oreginal: **LOW — password/confirm field errors are never cleared while typing.**

`onEmailChange` clears `fieldErrors.email` and `onAgreedChange` clears `fieldErrors.agree`, but `onPasswordChange={setPassword}` and `onConfirmChange={setConfirm}` clear nothing.

Submit with mismatched passwords → "Passwords do not match" → correct the confirm field → the red error stays put until the next submit.

The password field is worse, because `SignUpPane` gates its helper hint on `strength === "too-short" && !fieldErrors.password`. Once a length error is set, the *"Use at least 8 characters"* guidance never appears again for the rest of the session — and the stale "must be at least 8 characters" error sits underneath a now-valid 12-character password.

src/app/(auth)/verify-email/VerifyEmailContent.tsx
```diff
@@ -6,32 +6,64 @@ import { motion, AnimatePresence } from "framer-motion";
...
   const dest = resolveDestination(params.get("from"));
-  // /checkout is *enforced* (unlike advisory /account), so a manual "continue"
-  // there would just bounce off the checkout guard back to this page while
-  // unverified — and the verified case is already handled by the auto-redirect
-  // below. So the manual CTA only makes sense for the advisory destination.
-  const showManualContinue = !dest.startsWith("/checkout");
+  // A destination that requires verification would just bounce off its own guard back to
+  // this page while unverified — and the verified case is already handled by the
+  // auto-redirect below. So the manual CTA only makes sense for advisory destinations.
+  const showManualContinue = !requiresVerification(dest);
```
Oreginal, line 49: **LOW — the manual continue button is hardcoded to "Continue to My Account" for an arbitrary destination.**

`showManualContinue = !requiresVerification(dest)` renders the button for *any* non-checkout destination, and its `onClick={() => navOnce(dest)}` goes wherever `dest` points.

`SignupContent` pushes `withFrom("/verify-email", destination)` with whatever destination was resolved, so `/verify-email?from=%2Fstore` is reachable — and there the button says **"Continue to My Account"** while navigating to `/store`.

The label should follow `dest` rather than assume it.

src/context/AuthContext.tsx
```diff
@@ -77,70 +127,116 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+      return { displayNameSet, verificationEmailSent };
     },
     []
   );
 
   const socialSignIn = useCallback(async (provider: FirebaseAuthProvider) => {
     try {
-      await signInWithPopup(auth, provider);
-      setAuthCookie();
+      const result = await signInWithPopup(auth, provider);
```
Oreginal, line 225: **LOW — social sign-in records neither persistence nor the "remember me" marker, so the hint cookie can outlive the session.**

`signIn` carefully pairs `markSessionOnly(!remember)` with `setPersistence(...)` so the cookie lifetime and the Firebase session cannot drift. `socialSignIn` does neither — it just calls `setSessionHint()` and inherits whatever persistence the Auth instance was last left in.

Combined with `signOut` → `clearLocalUserState()`, which clears the `hb-session-only` marker but leaves persistence untouched, this drifts. All in one tab, no reload:

1. Sign in with **Remember me off** → persistence `browserSessionPersistence`, marker set.
2. Sign out → marker cleared, **persistence still session-scoped**.
3. Sign in with Google → `setSessionHint()` sees `isSessionOnly() === false` and writes the **30-day** cookie.

Firebase's session dies when the browser closes, but the hint survives a month. Next launch, `/account` renders behind a dead hint and the client bounces to `/login?…&session=expired` — precisely the stale-hint drift this PR set out to eliminate.

An explicit `markSessionOnly(false)` + `setPersistence(auth, browserLocalPersistence)` before the popup makes social sign-in state its intent instead of inheriting one.

src/app/(auth)/(card)/login/LoginContent.tsx
```diff
@@ -1,52 +1,111 @@
...
-    router.push(dest);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [user, authLoading]);
+    if (authLoading || !user || navigated.current) return;
+    navigated.current = true;
+    clearDestination();
+    router.push(postAuthDestination(destination, user.emailVerified));
+  }, [user, authLoading, destination, router]);
 
   const handleSubmit = async (e: React.FormEvent) => {
```
Oreginal, line 110: **LOW — conflict notices are not cleared when the user switches path.**

`handleSubmit` resets `error` but not `popupConflict`, so after a Google credential conflict the *"That email already has an account created with a different sign-in method"* panel stays pinned under the social buttons through a subsequent email/password attempt — including a failed one, where it sits next to an unrelated error.

The mirror case is in `SignupContent.tsx` `handleSocial`, which resets `popupConflict` but not `emailAlreadyInUse` — so the *"You already have an account with this email"* notice survives a switch to the Google button.

Each path should clear the other path's notice.

Oreginal (commented):
## Code review — PR #76

Reviewed the full 24-file diff (commit `8e13192`) against `main`, read the surrounding files, and ran `npx tsc --noEmit` — clean. Six findings below: one medium, five low. No blockers.

### The security core holds up

I went at `resolveDestination` adversarially and could not construct an input that escapes same-origin. It rejects `//`, any backslash, C0/DEL control characters and whitespace on the raw value, **re-checks the percent-decoded pathname** (so `/%09//evil.com` and `/%5C%5Cevil.com` both fall back), and validates the WHATWG-parsed origin. `withFrom` / `saveDestination` / `readDestination` all re-validate internally, so the `sessionStorage` path is covered rather than trusted.

The `signUp` partial-failure contract — *nothing after `createUserWithEmailAndPassword` may reject* — is correctly implemented, and it fixes a real bug where a cosmetic `updateProfile` failure told the user their sign-up failed on an account that already existed. The enumeration-safe message trio genuinely collapses to one string, and the legacy `?from=checkout` / `AUTH_COOKIE` / `getAuthErrorMessage`-from-toast references are all gone.

### Theme in the findings

Five of the six are the same class of bug: **state that is set but never unset.** A conflict notice, a "sent ✓" label, a field error, a persistence marker — each is written on one path and left for the next render to trip over. Individually minor; together they are why the auth screens accumulate stale UI across a session.

_Reviewed with Claude Code._

Oreginal:
## All six findings fixed — `bbe090b`

Pushed as [`bbe090b`](https://github.com/KatOre-Solutions/HeyBeautiful/commit/bbe090b324a3ce6a3e8cd402e04c8b052ad59176) (5 files, +58/−4).

| # | Finding | Fix |
| --- | --- | --- |
| 1 | MEDIUM — sign-out trips the "session expired" guard | `handleSignOut` claims `redirected.current` **before** `await signOut()`, so the expiry effect no-ops on the `user === null` render instead of racing `router.push("/")` |
| 2 | `resent` never reset | Added the `RESENT_NOTICE_MS` timeout effect, same shape as `VerifyEmailContent` — cleanup on unmount, no dangling timer |
| 3 | Password/confirm errors never cleared | `onPasswordChange` / `onConfirmChange` now clear their field errors; the password handler clears `confirm` too, since a length fix usually resolves the mismatch |
| 4 | Continue button hardcoded to "My Account" | Label derives from `dest` — "Continue to My Account" only when `dest === DEFAULT_DESTINATION`, otherwise plain "Continue" |
| 5 | Social sign-in inherits persistence | `socialSignIn` states its intent: `markSessionOnly(false)` + `setPersistence(auth, browserLocalPersistence)` before the popup |
| 6 | Conflict notices survive a path switch | `handleSubmit` clears `popupConflict`, `handleSocial` clears `emailAlreadyInUse` |

### Two notes on how these landed

**Finding 6 needed slightly more than reported.** `credentialConflict` is `popupConflict ?? redirectConflict`, so clearing `popupConflict` alone would still leave a *redirect*-path conflict panel pinned over an email/password attempt. `handleSubmit` now also calls `clearRedirectError()`, matching what `handleSocial` already did.

**Finding 5 closes the drift at both ends.** With `signIn` and `socialSignIn` now each stating their persistence explicitly, no sign-in path inherits whatever the Auth instance was last left in — so `signOut` clearing the `hb-session-only` marker without touching persistence can no longer strand a session-scoped login behind a 30-day hint cookie.

### Verification

- `npx tsc --noEmit` — clean
- `npx eslint` on all five modified files — clean
- `npm run build` — compiled successfully, 25/25 static pages generated

The security core called out in the review (`resolveDestination`, the `signUp` partial-failure contract, the enumeration-safe messages) is untouched by these fixes.

_Fixed with Claude Code._

PR #77, opened 2026-08-19
chore: add GitHub Actions CI pipeline and repair lint (#59)

### LINKED ISSUE #59: Add GitHub Actions CI pipeline (lint, typecheck, test, build)

Oreginal:
Closes #59. Also completes the remaining work for #52.

## Why the lint fix ships in this PR

`npm run lint` was **already failing on `main`** with two `react-hooks/rules-of-hooks` errors:

```
src/lib/shopify.ts
  238:7  error  React Hook "usePlaceholderCatalogue" is called in function "getProducts"…
  273:7  error  React Hook "usePlaceholderCatalogue" is called in function "getProductBySlug"…
```

`usePlaceholderCatalogue()` is a plain server-side boolean predicate, not a hook — the lowercase `use` prefix is the entire problem. It is easy to miss because linting individual files passes; only a whole-repo `eslint .` surfaces it.

CI runs `npm run lint`, so landing the workflow without this fix would have shipped a **permanently red pipeline on `main`**. Hence one PR.

Fixed by renaming to `placeholderCatalogueEnabled` — not by disabling the rule, adding disable comments, or scoping an override to `src/lib`. The rule still catches genuine hook bugs elsewhere.

## What's here

| File | Purpose |
| --- | --- |
| `.github/workflows/ci.yml` | One sequential job, cheapest-first: **lint → typecheck → test → build**. Push + PR on `main`/`master`. |
| `.nvmrc` | Node `22`, single-sourced |
| `package.json` | `typecheck` script (`tsc --noEmit`) + `engines.node: 22.x` |
| `.github/dependabot.yml` | Weekly npm + github-actions updates, targeting `main` |
| `src/lib/shopify.ts`, `src/lib/placeholder-products.ts` | The rename |
| `CLAUDE.md` | Verified stale reference: `toCartItem`/`defaultVariant`/`isSoldOut` are in `src/lib/product.ts`, not `shopify.ts` |

Notes on the workflow:

- `npm ci`, never `npm install` — the lockfile is committed and must be obeyed exactly.
- `.next/cache` keyed on the lockfile **plus** `src/**`, `public/**` and build config. Plain globs only: `hashFiles()` uses `@actions/glob`, which has no brace expansion, so `src/**/*.{ts,tsx}` would silently match nothing.
- `npm test --if-present` is a deliberate no-op — no runner exists yet, and CI picks tests up automatically once one does. No placeholder test added.
- The build runs **with no secrets**. Verified: `isConfigured()` returns false, the catalogue falls back to the non-purchasable "Coming Soon" tiles, and since those carry no `slug`, `generateStaticParams` prerenders no product detail pages.

## Verification

| Check | Result |
| --- | --- |
| `npm run lint` | exit **0** (was exit 1 on `main`) |
| `npm run typecheck` | exit **0** |
| `npm run build` with `.env.local` moved aside | exit **0** — 11 static pages, `/store/[slug]` prerenders none, as designed |
| `npm run build` normal | exit **0** — 25 static pages |
| Both YAML files | parsed; job name is byte-exact `Lint · Typecheck · Test · Build` |

## Follow-up (needs admin, not code)

Once this run reports, add **`Lint · Typecheck · Test · Build`** as a required status check on `main` and `master`. GitHub only offers a check for selection after it has reported at least once.

⚠️ Netlify has no `netlify.toml`, so it will now read `.nvmrc` as well. Pinning production and CI to the same Node major is intentional — worth watching the first `master` deploy after this lands.


.github/dependabot.yml
```diff
@@ -0,0 +1,32 @@
...
+# Updates open PRs against main, never master — master only ever receives a
+# promotion PR from main, per the branching strategy in CLAUDE.md.
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
+    target-branch: main
```
Oreginal: **MEDIUM — `target-branch` silently disables security updates.** ✅ documented in `b56ca5a`

The repository's default branch is `master` (confirmed via `gh repo view`). Dependabot **security** updates always target the default branch and cannot be retargeted — and per GitHub's docs, setting `target-branch` makes a config **version-updates only**.

Concrete effect: a critical CVE lands in `next` or `firebase` and **no security PR is ever opened**. You'd only pick it up whenever the weekly version run happens to bump past it.

**But removing `target-branch` is not the fix** — it would be actively worse. Dependabot PRs would then target `master`, where the `Only main may PR into master` required check fails *by design* for any head branch that isn't `main`. Every Dependabot PR would be permanently unmergeable.

The real fix is to make **`main` the repository default branch**. It *is* the integration branch; `master` is a deploy target. Then `target-branch` can be deleted from both entries and security updates start working.

That's a repo setting, not something this file can do — so `b56ca5a` documents the constraint and its cost in the file rather than shipping it silently.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+# A new push to the same branch makes the in-flight run obsolete.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
```
Oreginal, line 34: **LOW — no `permissions:` block, so `GITHUB_TOKEN` inherits the default.** ✅ fixed in `b56ca5a`

On older repositories that default is `contents: write`. This job only reads code — but `npm ci` executes third-party install scripts (`sharp` ships one) with that token in the environment. A compromised transitive dependency would find a push-capable token sitting there.

Added at workflow level:

```yaml
permissions:
  contents: read
```

Oreginal (commented):
## Code review — PR #77

Five findings: one medium, four low. **Four are fixed in `b56ca5a`; one I'm declining with evidence.** Details inline.

### Verified sound before reviewing

- **The rename is complete** — no `usePlaceholderCatalogue` references remain anywhere outside the generated `graphify-out/` artifact. Pure rename, no behaviour change.
- **The `CLAUDE.md` correction is accurate** — `defaultVariant` / `isSoldOut` / `toCartItem` are exported from `src/lib/product.ts`; the three fetchers are still in `shopify.ts`.
- **`typecheck` passes on a *fresh* checkout** — the risky case the PR body couldn't cover. `next-env.d.ts` and `.next/types/**` are both absent in CI yet listed in `tsconfig.json`'s `include`. Re-run with `next-env.d.ts` moved aside and `.next` removed: exit 0.
- **Building with no secrets is safe** — `src/lib/firebase.ts` has explicit `|| "missing-…"` fallbacks so `getAuth` can't throw during prerender.
- **All three action majors are real** — `checkout@v7`, `setup-node@v7`, `cache@v6`.

_Reviewed with Claude Code._

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+# A new push to the same branch makes the in-flight run obsolete.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
```
Oreginal: **LOW — `cancel-in-progress` also applies to `push`, leaving production commits unverified.** ✅ fixed in `b56ca5a`

The concurrency group keys on `github.ref`, and `cancel-in-progress: true` applied to the `push` trigger on `main`/`master`, not just PRs.

If two PRs merge into `master` within a few minutes, the first merge commit's run is cancelled and **that commit ends up with no CI verdict at all** — on the branch Netlify deploys from.

Now scoped so PR churn is still cancelled but branch pushes always finish:

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+      # cache; the restore-keys line still gives a warm start from the previous
+      # build whenever only sources moved.
+      - name: Restore Next.js build cache
+        uses: actions/cache@v6
+        with:
+          path: .next/cache
+          # Plain glob patterns only — hashFiles() uses @actions/glob, which has no
+          # brace expansion, so 'src/**/*.{ts,tsx}' would match nothing and quietly
+          # weaken the key. 'src/**' is broad on purpose.
+          key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('src/**', 'public/**', 'next.config.ts', 'tailwind.config.ts', 'postcss.config.mjs', 'tsconfig.json') }}
```
Oreginal: **LOW — `**/package-lock.json` is evaluated after `npm ci`, so it walks `node_modules`.** ✅ fixed in `b56ca5a`

This cache step sits *after* the install step, so the `**` glob traverses the freshly populated `node_modules`. Verified zero nested lockfiles there today — but the first dependency that ships one moves the cache key for reasons unrelated to this project, and every restore silently misses. It also makes `hashFiles` walk a large tree for no reason.

The lockfile is at the repo root, so it's now matched literally: `hashFiles('package-lock.json')`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+          restore-keys: |
+            ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-
```
Oreginal: **Declining this one — the source hash stays.**

The suggestion was to drop `hashFiles('src/**', 'public/**', …)` from the primary key and rely on the lockfile alone, on the grounds that `restore-keys` already recovers the previous cache and the source component just burns quota.

Two problems with that:

**1. It would freeze the cache permanently.** `actions/cache` does not overwrite an existing key. With a lockfile-only key, the very first run saves `.next/cache` and every subsequent run hits that exact key — so the cache is never updated again and goes stale forever. The source hash is precisely what triggers a fresh *save*. This is also why GitHub's own documented Next.js recipe includes a source hash alongside the lockfile.

**2. The quota premise is off by four orders of magnitude.** From the actual run log:

```
Restore Next.js build cache   Cache Size: ~0 MB (38870 B)
```

**38 KB**, not "multi-hundred-MB". The 193 MB entry in the same log is `setup-node`'s npm cache, which is keyed independently on the lockfile and is unaffected by this key. There is no LRU-eviction pressure to relieve here.

I've added a comment recording both reasons so this doesn't get "simplified" later.

Oreginal:
## Dependabot finding resolved at the root — `main` is now the default branch

Rather than living with the documented workaround, the underlying cause is fixed: **the repository default branch is now `main`** (was `master`).

That makes `target-branch` unnecessary *and* undesirable, so it's removed from both entries in `75434c5`. Dependabot now raises PRs against `main` because that's the default — which is where they belong under the promotion flow — and because no `target-branch` is set, **security updates work**: a critical CVE will open a security PR instead of waiting for the weekly version run to happen past it.

Verified after the switch:

| | |
| --- | --- |
| Default branch | `main` |
| `main` protection | 1 review, strict, `["Lint · Typecheck · Test · Build"]` |
| `master` protection | 1 review, strict, `["Only main may PR into master", "Lint · Typecheck · Test · Build"]` |
| Open PRs | #74, #77, #79 all still target `main`, unaffected |
| CI on `75434c5` | ✅ green |

⚠️ **One thing to confirm outside GitHub:** Netlify has its own *production branch* setting and there's no `netlify.toml` in the repo. Changing the GitHub default does **not** change Netlify's production branch, so deploys should still come from `master` — worth a glance at the Netlify site settings to be sure it wasn't following the repo default.

PR #79, opened 2026-08-19
fix: harden the restrict-pr-source workflow (#78)

### LINKED ISSUE #78: Harden restrict-pr-source workflow: untrusted input in run:, and fork branches named main pass the check
> **Correction (severity revised down).** This was first reported as a shell-injection *bypass* of the production guard. Testing does not support that: the naive payloads fail **closed**, not open. Details below. The fork hole in part 2 is the substantive issue; part 1 is hardening.

`.github/workflows/restrict-pr-source.yml` is live on `master`, and its job `Only main may PR into master` is a **required status check** there — the only thing enforcing "production may only be reached from `main`".

## Part 1 — untrusted input interpolated into `run:` (hardening)

```yaml
run: |
  if [ "${{ github.head_ref }}" != "main" ]; then
```

`${{ }}` is substituted as **text** before the shell parses the line, and branch names may contain `"`, `;`, `$`, backtick and `|`. This is the `actions/command-injection` anti-pattern that CodeQL flags and that GitHub's own hardening guidance tells you to avoid.

**What it does not do, contrary to the original report:** it does not yield a working bypass here. Bash parses the whole `if…fi` compound before executing any of it, so a malformed branch name is a syntax error — the step exits non-zero and the check **fails closed**. Measured against the real script:

| Branch name | Exit | Result |
| --- | --- | --- |
| `x"; exit 0; #` | 2 | fails closed |
| `x";exit${IFS}0;#` | 2 | fails closed |
| `x";exit${IFS}0;if${IFS}true;then${IFS}#` | 2 | fails closed |
| `feature-branch` | 1 | correctly blocked |
| `main` | 0 | correctly allowed |

Git's ref-name rules help here too — no spaces, no `[`, no `:` — which rules out the usual ways of reopening a `test` or supplying a no-op. I could not construct a payload that both parses and reaches a green `exit 0`.

So: worth fixing as hardening and to keep CodeQL quiet, **not** an open door to production.

## Part 2 — a fork's `main` passes the check (real)

`github.head_ref` is only the branch *name*. It cannot distinguish this repository's `main` from a fork's `main`, so anyone can fork, push a branch named `main`, and open a PR straight into production `master` — and the check goes green. No exploit construction needed; it simply works.

This is the part that actually matters.

## Fix

1. Pass untrusted values through `env:` and quote them on use.
2. Assert `github.event.pull_request.head.repo.full_name == github.repository`.
3. `permissions: {}` — the job only compares strings and needs no token.

## Notes

- The vulnerable copy is already on `master`; `main` does not have the file.
- PR #74 adds the same file, byte-identical and unfixed. Whichever lands second conflicts — the fixed version must win.
- Fixing on `main` does not protect production until a `main` → `master` promotion carries it across.

Oreginal:
Closes #78.

Adds `.github/workflows/restrict-pr-source.yml` to `main` in a hardened form. `main` doesn't have this file today; `master` does, unhardened.

## ⚠️ Correction to the original report

I first reported the `${{ github.head_ref }}` interpolation as a HIGH-severity **bypass** of the production guard, on PR #74. **That was wrong, and I've corrected it there and on #78.** Testing doesn't support it — the payloads fail **closed**, not open.

Bash parses the entire `if…fi` compound before executing any of it, so a malformed branch name is a syntax error: nothing runs, the step exits non-zero, the check fails, the PR is blocked. Measured against the real script:

| Branch name | Exit | Result |
| --- | --- | --- |
| `x"; exit 0; #` | 2 | fails closed |
| `x";exit${IFS}0;#` | 2 | fails closed |
| `x";exit${IFS}0;if${IFS}true;then${IFS}#` | 2 | fails closed |
| `feature-branch` | 1 | correctly blocked |
| `main` | 0 | correctly allowed |

Git's ref-name rules (no spaces, no `[`, no `:`) rule out the usual ways of reopening a `test` or supplying a no-op. I could not construct a payload that both parses and reaches a green `exit 0`.

## What's actually here

**1. Untrusted input out of `run:` — hardening.** Values now go through `env:` and are quoted on use. `${{ }}` is substituted as *text* before the shell parses the line, which is the `actions/command-injection` anti-pattern CodeQL flags and GitHub's hardening guide warns against. Worth fixing on principle; not a live hole, per above.

**2. Fork branches named `main` passed the check — this one is real.** `github.head_ref` is only a branch *name*, so it cannot tell this repo's `main` from a fork's `main`. Anyone could fork, push a branch called `main`, and open a PR straight into production `master` with the check green. No exploit construction needed — it simply worked. The job now asserts `head.repo.full_name == github.repository` first.

**3. `permissions: {}`** — the job only compares two strings and needs no token.

The job name **`Only main may PR into master` is unchanged**, and must stay that way: it's the required status check on `master`, and renaming it would silently detach the protection rule so that everything passes.

## Merge notes

- This does **not** protect production until a `main` → `master` promotion carries it across. `master` keeps the unhardened copy until then.
- **PR #74 adds this same file, byte-identical and unfixed.** Whichever lands second will conflict; the hardened version must win.
- This PR is blocked until #77 merges — `main` now requires the `Lint · Typecheck · Test · Build` check, and `ci.yml` isn't on `main` yet, so the check can never report here.

### DIFF
--- a/.github/workflows/restrict-pr-source.yml
+++ b/.github/workflows/restrict-pr-source.yml
@@ -1,0 +1,48 @@
+# Enforces the promotion flow from CLAUDE.md: production (`master`) may only ever
+# be reached from `main`. This job is a REQUIRED STATUS CHECK on master, so it is
+# the thing standing between an arbitrary branch and production.
+name: Restrict PR source branch
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+    branches:
+      - master
+
+# The job only compares two strings — it needs no repository access at all.
+permissions: {}
+
+jobs:
+  check-source-branch:
+    # Do NOT rename: this exact string is the required status check configured on
+    # master. Renaming it detaches the protection rule, which then passes silently.
+    name: Only main may PR into master
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify head branch
+        # Every untrusted value is passed as an environment variable and quoted on
+        # use. `${{ }}` is substituted as TEXT before the shell parses the line, so
+        # interpolating a branch name directly into `run:` is a command injection:
+        # branch names may contain ", ;, $, backtick and |. A PR from a branch named
+        #   x"; exit 0; #
+        # expanded to `if [ "x"; exit 0; #" != "main" ]; then`, where `[` failed as a
+        # syntax error, `exit 0` ran, and the step reported SUCCESS — defeating the
+        # only guard on production. (#78)
+        env:
+          HEAD_REF: ${{ github.head_ref }}
+          HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name }}
+          THIS_REPO: ${{ github.repository }}
+        run: |
+          # head_ref is only a branch NAME, so on its own it cannot tell this repo's
+          # `main` from a fork's `main`. Check the repository first.
+          if [ "$HEAD_REPO" != "$THIS_REPO" ]; then
+            echo "PRs into master must come from $THIS_REPO — this one is from the fork '$HEAD_REPO'."
+            exit 1
+          fi
+
+          if [ "$HEAD_REF" != "main" ]; then
+            echo "PRs into master are only allowed from the 'main' branch. This PR is from '$HEAD_REF'."
+            exit 1
+          fi
+
+          echo "OK: PR is from main."
--- a/.github/workflows/restrict-pr-source.yml
+++ b/.github/workflows/restrict-pr-source.yml
@@ -1,6 +1,28 @@
 # Enforces the promotion flow from CLAUDE.md: production (`master`) may only ever
 # be reached from `main`. This job is a REQUIRED STATUS CHECK on master, so it is
 # the thing standing between an arbitrary branch and production.
+#
+# ⚠️ KNOWN LIMITATION — this guard can be NEUTERED by the PR it is guarding.
+#
+# `on: pull_request` resolves workflow files from the PR's MERGE COMMIT, not from
+# the base branch. Proven in this repo: ci.yml did not exist on main at e39c1cc,
+# yet it ran twice as `pull_request` on PR #77 — the PR that introduced it.
+#
+# So a PR into master can edit THIS FILE in the same branch — `exit 0`, or a
+# job-level `if: false`, which reports as "skipped" and satisfies a required
+# check — and the check named below goes green while the PR is not from main.
+# The repo is public, so anyone can fork and try it. Merging still needs a human
+# approving review, but the signal an approver relies on would be lying.
+#
+# The fix is `pull_request_target`, which resolves from the base branch and is
+# safe here precisely because this job checks nothing out and holds no token.
+# It CANNOT be applied in one step: pull_request_target eligibility is read from
+# the base, and pull_request eligibility from the merge commit, so flipping the
+# trigger means NEITHER fires on the promotion PR — the required check never
+# reports and production promotion is blocked permanently. Migrating needs two
+# promotion cycles (add pull_request_target alongside, land it on master, then
+# drop pull_request). Tracked separately rather than risking the deploy path
+# inside a hardening PR.
 name: Restrict PR source branch
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/restrict-pr-source.yml
```diff
@@ -0,0 +1,48 @@
+# Enforces the promotion flow from CLAUDE.md: production (`master`) may only ever
+# be reached from `main`. This job is a REQUIRED STATUS CHECK on master, so it is
+# the thing standing between an arbitrary branch and production.
+name: Restrict PR source branch
+
+on:
```
Oreginal, line 6: **MEDIUM — the guard can be neutered by the PR it is guarding.** 📋 tracked in #86, deliberately not fixed here

`on: pull_request` resolves workflow files from the PR's **merge commit**, not from the base branch. I proved this in this repo rather than assuming it: `ci.yml` was **absent from `main` at `e39c1cc`**, yet it ran twice as `pull_request` on PR #77 — the PR that introduced it.

So a PR into `master` can edit *this file* in the same branch — `exit 0`, or a job-level `if: false`, which reports as **skipped and satisfies a required check** — and the check goes green while the PR is not from `main`. The repo is public, so anyone can fork and try it.

Not a full break-glass: merging still needs a human approving review, and first-time contributors need a maintainer to click "Approve and run". But the automated signal an approver leans on would be lying, which is the whole point of the check.

**Why I didn't just fix it.** `pull_request_target` is the right answer — it resolves from the base branch, and is safe here precisely because this job checks nothing out and holds no token. But it cannot be a one-line change:

| Event | Read from | `main` would have | `master` has | Fires? |
| --- | --- | --- | --- | --- |
| `pull_request` | merge commit | `pull_request_target` | — | ❌ |
| `pull_request_target` | base branch | — | `pull_request` | ❌ |

Neither fires. The required check never reports and **`main` → `master` promotion is blocked permanently** until an admin strips the protection. Flipping this trigger inside a hardening PR would take out the deploy path.

It needs a two-promotion-cycle migration (add alongside → land on master → drop `pull_request`), which is written up in **#86**. `7d1fd05` records the limitation in the file header so nobody assumes this guard is stronger than it is.

.github/workflows/restrict-pr-source.yml
```diff
@@ -0,0 +1,48 @@
...
+    # Do NOT rename: this exact string is the required status check configured on
+    # master. Renaming it detaches the protection rule, which then passes silently.
+    name: Only main may PR into master
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify head branch
+        # Every untrusted value is passed as an environment variable and quoted on
+        # use. `${{ }}` is substituted as TEXT before the shell parses the line, so
+        # interpolating a branch name directly into `run:` is a command injection:
+        # branch names may contain ", ;, $, backtick and |. A PR from a branch named
```
Oreginal, line 26: **LOW — this comment documents an exploit I already retracted.** ✅ fixed in `7d1fd05`

Good catch, and squarely my fault. Lines 26–30 still asserted that `x"; exit 0; #` expanded such that "`exit 0` ran, and the step reported SUCCESS". That's false — I corrected it in the PR body and on #78, but left the file comment saying the opposite.

Bash parses the whole `if…fi` compound before executing any of it, so a malformed branch name is a syntax error: the shell exits 2 without running a single command. It fails **closed**.

The damage of leaving it: it tells a future maintainer that the unhardened copy still sitting on `master` is a live, exploitable route into production — which could trigger an unwarranted emergency response — while simultaneously undercutting the accurate justification right above it (untrusted input in `run:` is the `actions/command-injection` pattern and is not *reliably* fail-closed).

Rewritten to state the measured behaviour, and to say plainly that this half is defence in depth while the same-repo assertion below is the half that closed a real hole.

Oreginal (commented):
## Code review — PR #79

Two findings, both real. **One fixed in `7d1fd05`; one is a deeper design flaw I've documented here and tracked in #86 rather than fixing inline — because the obvious fix would block production promotion permanently.**

### Verified sound

The hardening itself holds up. `env:` + quoted expansion removes the text-substitution pattern; the same-repo assertion genuinely closes the fork-named-`main` hole; `permissions: {}` is right for a job that only compares strings; and the required-check job name `Only main may PR into master` is byte-identical to `origin/master`'s copy, so branch protection stays attached. Every unhappy path still exits 1 — empty `HEAD_REF`, and `head.repo` null on a deleted fork.

Also checked and cleared: `types: [… edited]` does cover the "open into `main`, then re-target to `master`" path; the committed blob is LF despite local `core.autocrlf=true`, so no CRLF-in-`run:` problem; and `HEAD_REF` can't smuggle Actions workflow commands because `:` is illegal in a git refname.

_Reviewed with Claude Code._

PR #85, opened 2026-08-19
fix: stop the wishlist opening a second bag row per product (#63)

### LINKED ISSUE #63: Wishlist and cart disagree on item id, producing duplicate bag rows

Oreginal:
Closes #63.

## The collision

Two id schemes met in the middle:

| Surface | Keys by | Example |
| --- | --- | --- |
| Cart | **variant** (so two sizes are separate rows) | `product:1#4567` |
| Wishlist | **product** (so the heart fills) | `product:1` |

`WishlistSidebar` handed a `WishlistProduct` straight to `addItem`, so quick-add from a card produced `product:1#4567` while move-to-bag produced `product:1` — the same product sitting in the bag twice, with independent quantities and an "in cart" state that disagreed between surfaces.

## Why neither obvious fix works

The issue suggested either making the wishlist variant-keyed, or routing move-to-bag through `toCartItem()`. Both dead-end:

- **Variant-keyed wishlist** breaks the hearts. Both `isWishlisted(...)` call sites pass the bare `product.id`, so a variant-keyed entry would never match and the heart would stop filling on the card and the detail page.
- **Calling `toCartItem()` on the way out** isn't possible — it needs a `ShopifyProduct`, and a `WishlistProduct` carries no variant data at all.

## What this does instead

The entry keeps its product-level identity and **carries the bag line `toCartItem()` built for it at wish time**. `WishlistSidebar` moves *that* across.

- The **card** captures the default variant.
- The **detail page** captures the variant the shopper is actually looking at — so wishlisting the 60-serving size and moving it over adds the 60-serving line, not the 30.

`cartLine` is optional and falls back to the product fields, which is the old behaviour and correct for the two cases that need it: **bundles** (no variants) and **entries persisted before this shipped** (no migration required).

`toCartItem`'s return type is now named `CartLine`, so the shape is carried verbatim rather than rebuilt — cart-id construction stays in exactly one place.

**Both bag paths were affected.** The issue only mentions `handleMoveToBag`; `handleShopAll` had the same defect and is fixed too.

## Verification

Exercised against the **real** `toCartItem` via `node --experimental-strip-types` (the module is import-free, so it loads directly — no reimplementation):

```
BEFORE fix -> rows: 2   product:1#4567 x1 | product:1 x1
AFTER  fix -> rows: 1   product:1#4567 x2
variant B  -> rows: 2   product:1#4567 x1 | product:1#8910 x1
bundle     -> rows: 1   bundle:glow x1
```

The last two are the negative cases that matter: a genuinely different variant must still get its own row, and a `cartLine`-less entry must still work.

Also: `npm run lint` ✅ · `npm run typecheck` ✅ · `npm run build` ✅

`CLAUDE.md` now records which surface owns the variant suffix, as the issue asked.

### DIFF
--- a/src/app/store/[slug]/ProductDetailContent.tsx
+++ b/src/app/store/[slug]/ProductDetailContent.tsx
@@ -73,6 +73,10 @@ export default function ProductDetailContent({
       category: product.category,
       price: product.price,
       image: product.image,
+      // The variant the shopper is actually looking at — same line the Add to
+      // Bag button above would build, so moving it over from the wishlist adds
+      // to that row rather than opening a second one (#63).
+      cartLine: toCartItem(product, variant),
     });
   };
 
--- a/src/components/WishlistSidebar.tsx
+++ b/src/components/WishlistSidebar.tsx
@@ -15,12 +15,23 @@ export default function WishlistSidebar() {
     useWishlist();
   const { addItem, setCartOpen } = useCart();
 
-  // The wishlist keys entries by PRODUCT while the cart keys lines by VARIANT,
-  // so a saved item can't be handed to the cart as-is — that produced a bare
+  // The wishlist keys entries by PRODUCT while the cart keys lines by VARIANT, so
+  // a saved item can't be handed to the cart as-is — that produced a bare
   // `product:1` row alongside the `product:1#4567` the card's quick-add builds,
-  // i.e. the same product twice with independent quantities (#63). `cartLine`
-  // was captured by toCartItem() at wish time; fall back to the product fields
-  // for entries saved before it existed, and for bundles, which have no variants.
+  // i.e. the same product twice with independent quantities (#63). `cartLine` is
+  // captured by toCartItem() at wish time.
+  //
+  // This is also what the row RENDERS, not just what it adds. The two must come
+  // from one source: `item.price` is priceRange.minVariantPrice and `item.name`
+  // has no variant suffix, while cartLine carries the selected variant's price
+  // and label. Rendering the product fields while adding the variant line let the
+  // drawer read "Plant Protein Luxe · R 450" and then drop
+  // "Plant Protein Luxe — 60 servings · R 750" into the bag. The card path
+  // diverges too whenever defaultVariant() — first *available*, not cheapest —
+  // isn't the min-price variant.
+  //
+  // Entries restored without a cartLine are dropped in WishlistContext, so the
+  // fallback here is a type-level total rather than a live path.
   const toBagLine = (item: WishlistProduct) => {
     const { cartLine, ...product } = item;
     return cartLine ?? product;
@@ -116,7 +127,10 @@ export default function WishlistSidebar() {
       ) : (
         <div className="p-5 flex flex-col gap-3">
           <AnimatePresence mode="popLayout">
-            {items.map((item) => (
+            {items.map((item) => {
+              // Show exactly what Move to bag will add — see toBagLine.
+              const line = toBagLine(item);
+              return (
               <motion.div
                 key={item.id}
                 layout
@@ -138,7 +152,7 @@ export default function WishlistSidebar() {
                 >
                   <Image
                     src={item.image}
-                    alt={item.name}
+                    alt={line.name}
                     fill
                     className="object-cover"
                     sizes="60px"
@@ -151,7 +165,7 @@ export default function WishlistSidebar() {
                     className="label-caps mb-1"
                     style={{ color: "rgba(201,151,122,0.8)", fontSize: "8px" }}
                   >
-                    {item.category}
+                    {line.category}
                   </p>
                   <p
                     className="leading-tight mb-2 truncate"
@@ -162,7 +176,7 @@ export default function WishlistSidebar() {
                       fontWeight: 500,
                     }}
                   >
-                    {item.name}
+                    {line.name}
                   </p>
                   <p
                     style={{
@@ -172,7 +186,7 @@ export default function WishlistSidebar() {
                       fontWeight: 400,
                     }}
                   >
-                    {formatPrice(item.price)}
+                    {formatPrice(line.price)}
                   </p>
                 </div>
 
@@ -187,7 +201,7 @@ export default function WishlistSidebar() {
                       background: "rgba(201,151,122,0.1)",
                       border: "1px solid rgba(201,151,122,0.25)",
                     }}
-                    aria-label={`Move ${item.name} to bag`}
+                    aria-label={`Move ${line.name} to bag`}
                   >
                     <ShoppingBag size={12} className="text-rose-gold" />
                   </motion.button>
@@ -201,13 +215,14 @@ export default function WishlistSidebar() {
                       background: "rgba(239,68,68,0.07)",
                       border: "1px solid rgba(239,68,68,0.18)",
                     }}
-                    aria-label={`Remove ${item.name} from wishlist`}
+                    aria-label={`Remove ${line.name} from wishlist`}
                   >
                     <X size={12} style={{ color: "rgba(239,68,68,0.7)" }} />
                   </motion.button>
                 </div>
               </motion.div>
-            ))}
+              );
+            })}
           </AnimatePresence>
         </div>
       )}
--- a/src/components/WishlistSidebar.tsx
+++ b/src/components/WishlistSidebar.tsx
@@ -15,10 +15,21 @@ export default function WishlistSidebar() {
     useWishlist();
   const { addItem, setCartOpen } = useCart();
 
+  // The wishlist keys entries by PRODUCT while the cart keys lines by VARIANT,
+  // so a saved item can't be handed to the cart as-is — that produced a bare
+  // `product:1` row alongside the `product:1#4567` the card's quick-add builds,
+  // i.e. the same product twice with independent quantities (#63). `cartLine`
+  // was captured by toCartItem() at wish time; fall back to the product fields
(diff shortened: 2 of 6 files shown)


src/context/WishlistContext.tsx
```diff
@@ -42,9 +66,23 @@ export function WishlistProvider({ children }: { children: ReactNode }) {
...
+        // a stale entry cannot be repaired, only dropped.
+        //
+        // Every current write sets cartLine, including single-variant products
+        // (toCartItem returns the bare product id there, but still returns a
+        // line), so `cartLine == null` identifies legacy entries exactly.
+        // Bundles never enter the wishlist — BundleCard has no wishlist control —
+        // so nothing legitimate is caught by this.
         const restored = (JSON.parse(raw) as WishlistProduct[]).filter(
-          (p) => typeof p.id === "string"
+          (p) => typeof p.id === "string" && p.cartLine != null
```
Oreginal, line 85: **MEDIUM — the fix never reached anyone who already had a wishlist.** ✅ fixed in `4bfa345`

Wishlists persist in `localStorage` indefinitely and nothing migrated them. An entry hearted before this ships is `{id: "product:1", …}` with no `cartLine` — so it falls through `toBagLine`'s fallback to the bare product id and **reproduces #63 exactly**:

> quick-add "Glow Collagen Blend" from the grid → `product:1#4567`, then Move to bag on the pre-existing saved copy → `product:1` → two rows, one product, independent quantities.

And my justification for that fallback was wrong. I wrote it existed for "bundles, which have no variants" — but `BundleCard` has no wishlist control (`addItem` only, no `Heart`/`toggleItem`), so **bundles never enter the wishlist**. The fallback was reachable only by the legacy case, i.e. only by the broken one.

Now filtered on restore. `cartLine == null` identifies legacy entries exactly, because every current write sets it — including single-variant products, where `toCartItem` returns the bare product id but still returns a *line*.

Dropping rather than repairing is deliberate: a stored entry doesn't carry the variant list, so there is nothing to rebuild the id from.

src/components/WishlistSidebar.tsx
```diff
@@ -15,10 +15,32 @@ export default function WishlistSidebar() {
...
+  // has no variant suffix, while cartLine carries the selected variant's price
+  // and label. Rendering the product fields while adding the variant line let the
+  // drawer read "Plant Protein Luxe · R 450" and then drop
+  // "Plant Protein Luxe — 60 servings · R 750" into the bag. The card path
+  // diverges too whenever defaultVariant() — first *available*, not cheapest —
+  // isn't the min-price variant.
+  //
+  // Entries restored without a cartLine are dropped in WishlistContext, so the
+  // fallback here is a type-level total rather than a live path.
+  const toBagLine = (item: WishlistProduct) => {
```
Oreginal, line 35: **MEDIUM — the drawer quoted a different figure from the one the bag received.** ✅ fixed in `4bfa345`

This was a regression the first pass introduced. The row rendered `item.name` and `formatPrice(item.price)` — product-level, where `price` is `priceRange.minVariantPrice` (`shopify.ts:194`) — while `toBagLine` now hands the cart `cartLine`, carrying the variant's label and price. **Before the first pass, the two agreed.**

It bites on the card path too, not just the detail page: `defaultVariant()` returns the first *available* variant, not the cheapest, so any product whose cheapest variant is sold out diverges.

Worked example, verified against the real `toCartItem`:

```
before  drawer: "Plant Protein Luxe"                R 450
        bag   : "Plant Protein Luxe — 60 servings" R 750
after   both  : "Plant Protein Luxe — 60 servings" R 750
```

Fixed at the root rather than by patching the two bindings: the row derives `const line = toBagLine(item)` and renders *that*. One source of truth, so display and action can't drift again.

src/app/store/[slug]/ProductDetailContent.tsx
```diff
@@ -73,6 +73,10 @@ export default function ProductDetailContent({
       category: product.category,
       price: product.price,
       image: product.image,
+      // The variant the shopper is actually looking at — same line the Add to
+      // Bag button above would build, so moving it over from the wishlist adds
+      // to that row rather than opening a second one (#63).
+      cartLine: toCartItem(product, variant),
```
Comment (Oreginal) [inline src/app/store/[slug]/ProductDetailContent.tsx:79]: **LOW — a captured `cartLine` is never refreshed.** 📋 tracked in #91

`handleWish` is a pure toggle keyed on `product.id`, so an entry that already exists is removed, never updated. Concretely: heart the product from a store card (captures the 30-serving line), then open the detail page and select 60 servings — the heart already renders filled, so clicking it *removes* the entry. The saved line stays 30-serving.

So this PR's stated behaviour — "wishlisting the 60-serving size and moving it over adds the 60-serving line" — holds only when the product isn't already wishlisted. **I've corrected that claim in the PR description.**

Not fixed here because the repair is a product decision, not a mechanical one: overwriting on click means the filled heart sometimes updates instead of un-hearting (contradicting what the icon communicates), and overwriting on variant-change means the dropdown silently rewrites saved data. #91 lays out the options.

src/components/WishlistSidebar.tsx
```diff
@@ -28,7 +50,7 @@ export default function WishlistSidebar() {
   // No auto-open of the cart — the "Added to bag" toast confirms it (matches
   // the product-card add behaviour).
   const handleMoveToBag = (item: WishlistProduct) => {
```
Oreginal, line 52: **LOW — move-to-bag has no sold-out guard.** 📋 tracked in #91

`WishlistHeart` renders for every non-placeholder product including sold-out ones (`ShopifyProductCard.tsx` gates only on `product.placeholder`), while both genuine add paths refuse — the card hides quick-add behind `purchasable`, and `handleAddToCart` returns early on `soldOut`. Move to bag and **Shop All Saved Items** don't.

So hearting a fully sold-out product stores `product:1#<sold-out-variant>` (via `defaultVariant()`'s `variants[0]` fallback) and Move to bag drops an unpurchasable line into the cart. Pre-existing for the product-level id, but this PR makes the line variant-specific.

Not fixed here because a `WishlistProduct` carries no availability field, and availability is *live* — anything hearted can sell out afterwards. A correct guard needs either a refetch at drawer-render time or an explicit snapshot flag, plus a decision on whether such a row should be visibly disabled or rejected later. #91.

Oreginal (commented):
## Code review — 4 findings

**Two fixed in `4bfa345`; two tracked in #91** because they need a UX decision rather than a mechanical fix.

The first pass had a real problem I should have caught: **it never reached anyone who already had a wishlist**, and it introduced a price/name mismatch that didn't exist before. Both are now fixed and covered by a behavioural test.

### Fixed

**1 — The fix never reached existing users (medium).** Wishlists live in `localStorage` indefinitely and nothing migrated them. An entry saved before this shipped has no `cartLine`, fell through to its bare `product:1`, and reproduced the duplicate bag row **verbatim**.

My justification for that fallback was also just wrong. I wrote that it existed for "bundles, which have no variants" — but `BundleCard` has no wishlist control at all, so bundles never enter the wishlist. The fallback was reachable **only** by the broken case it was papering over.

Legacy entries are now dropped on restore. They can't be repaired — a stored entry doesn't carry the variant list, so the variant id can't be rebuilt — only discarded.

**2 — The drawer quoted a different price from the one the bag received (medium).** This one was a regression *introduced* by the first pass: the row rendered `item.name`/`item.price` (product-level; `price` is `priceRange.minVariantPrice`) while `toBagLine` handed the cart the variant's name and price. Before the first pass they agreed.

Measured against the real `toCartItem`, with the cheapest variant sold out so `defaultVariant()` picks the pricier one:

```
before  drawer: "Plant Protein Luxe"                 R 450
        bag   : "Plant Protein Luxe — 60 servings"  R 750   <-- MISMATCH
after   both  : "Plant Protein Luxe — 60 servings"  R 750
```

The row now renders the same line `toBagLine` returns, so display and action **cannot** drift apart again.

### Verification

```
F2  drawer/bag agree ............ PASS
F1  legacy entry dropped ........ PASS  (1 -> 0 on restore)
    current entry survives ...... PASS  (product:2#222)
```

`npm run typecheck` ✅ · `npm run lint` ✅ · `npm run build` ✅

### Tracked in #91, not fixed here

Both are pre-existing in shape; this PR makes the entry variant-specific, which sharpens them. Both need a product decision, so guessing at it inside a bug fix would be the wrong call.

_Reviewed with Claude Code._

Issue #65, opened 2026-08-10
Bundles are hardcoded and have no Shopify equivalent


Oreginal:
Split out from #62, which only fixed the *currency symbol* on bundles (they rendered `$128` next to
Shopify products priced in rand).

## Current state

Products now come from Shopify, but `Bundle` / `bundles` remain hardcoded in `src/lib/products.ts` —
ids, names, prices, `originalPrice`, and the `savings` string (`"21% off"`) are all literals. So
bundle pricing can drift from the product pricing it's derived from, and nobody finds out.

`src/lib/products.ts` now exists solely for this.

## Options

- Model bundles as real Shopify products (simplest — they become ordinary catalog entries and the
  hardcoded module disappears).
- Keep them as a merchandising construct but derive price/savings from their component products at
  build time, so they can't drift.

This is a merchandising decision as much as a technical one.

Issue #91, opened 2026-08-20
Wishlist: stale variant on already-wishlisted products, and no sold-out guard on move-to-bag


Oreginal:
Two findings from the review of #85 that need a UX decision rather than a mechanical fix, so they were deliberately left out of that PR. Both are pre-existing in shape; #85 makes the wishlist entry variant-specific, which sharpens them.

## 1. The heart is a pure toggle, so a captured `cartLine` is never refreshed

`handleWish` in `src/app/store/[slug]/ProductDetailContent.tsx` calls `toggleItem({...})` keyed on `product.id`. Since `isWishlisted` matches on the product id, an entry that already exists is **removed**, never updated.

Scenario:

1. Shopper hearts "Plant Protein Luxe" from a store card → entry captures the default 30-serving line.
2. Shopper opens the detail page and selects **60 servings**.
3. The heart already renders filled, so clicking it *removes* the entry rather than re-capturing it.
4. The saved entry keeps the 30-serving `cartLine`, and Move to bag adds the 30-serving row.

So #85's stated behaviour — "wishlisting the 60-serving size and moving it over adds the 60-serving line" — holds only when the product is not already wishlisted.

**Why it wasn't just fixed:** the obvious repair is to overwrite the entry when the selected variant differs, but that means either (a) the filled heart sometimes updates instead of un-hearting, which contradicts what the icon communicates, or (b) merely changing the variant dropdown silently rewrites saved data the shopper didn't ask to change. Both are product decisions.

Options worth weighing:

- Keep the wishlist strictly product-level and drop the variant from the entry entirely — Move to bag would then always add the current default variant.
- Add an explicit "update saved item" affordance when the selected variant differs from the saved one.
- Let the heart overwrite silently and accept (b).

## 2. Move-to-bag has no sold-out guard

`WishlistHeart` renders for every non-placeholder product, including sold-out ones — `ShopifyProductCard.tsx` gates only on `product.placeholder`. Both genuine add paths refuse a sold-out product (the card hides quick-add behind `purchasable`; `handleAddToCart` returns early on `soldOut`), but `WishlistSidebar`'s Move to bag and **Shop All Saved Items** do not.

Scenario: a shopper hearts a fully sold-out product. `defaultVariant()` falls back to `variants[0]`, so the entry carries `product:1#<sold-out-variant>`, and Move to bag drops an unpurchasable line into the cart.

[...]

Issue #31, opened 2026-06-20
Integrate the payment gateway with Shopify (hosted checkout handoff)


KatlehoMadaba:
**This is the last thing between the storefront and revenue.** Everything else on the board is
polish, hardening, or configuration.

## Current state

`/checkout` renders a correct order summary and then stops. The purchase control is a static div:

```tsx
{/* Payment is not wired up yet — honest placeholder. */}
…
Secure payment coming soon
```

Confirmed by search across `src/`:

- `cartCreate`, `checkoutCreate`, `cartLinesAdd`, `checkoutUrl`, `mutation` → **zero hits.** The
  Shopify layer in `src/lib/shopify.ts` is read-only.
- `stripe|paypal|payfast|yoco|peach|ozow|paystack` across `src/`, `.env.example`, `package.json`
  → **zero hits.**

So a customer can browse, pick a variant, fill a bag, sign in, verify their email, reach checkout
— and cannot buy. #4 records this precisely: every box ticked except `[ ] Checkout handoff`.

## Recommended approach: hand off to Shopify's hosted checkout

Create a Shopify cart from the local bag and redirect the customer to the `checkoutUrl` Shopify
returns. Shopify then owns payment, PCI compliance, tax, shipping rates, discount codes, order
records, and the confirmation email.

**Why not a custom payment integration.** Building our own would mean owning PCI scope, tax and
shipping calculation, an order data model, refunds, and a fraud story — for a store this size
that is months of work and permanent liability, to arrive at a worse checkout than the one
Shopify already runs.

Two more things it buys us:

- **Price integrity.** Cart prices currently live in `localStorage` and are user-editable. Any
  self-hosted payment flow would have to re-price server-side before charging. Hosted checkout
  re-prices from variant ids automatically — the bug class disappears instead of needing defence.
- **It largely dissolves #57.** That issue asks for server-side `email_verified` enforcement on
  order/payment endpoints. With hosted checkout there is no self-hosted order endpoint to
  protect. The `/checkout` client gate stays what it always was — UX, not a security boundary —
  and #57 should be re-read and probably closed once this lands.

## Blocked by #92

`cartCreate` needs `merchandiseId: "gid://shopify/ProductVariant/<id>"`. The numeric variant id
is already encoded in the cart key (`product:1#42`) but isn't addressable as a field. #92 exposes
it and — importantly — forces the decision about **lines that have no variant at all**:

[...]

PR #93, opened 2026-08-25
feat: expose the Shopify variant id on the cart line (#92)

### LINKED ISSUE #92: Expose the Shopify variant id on the cart line (prerequisite for checkout)
Prerequisite for #31 (payment gateway) and the last box on #4. Filing separately because it's a
self-contained change with no Shopify admin dependency — it can land and be reviewed before the
checkout work starts.

## The situation

Shopify's `cartCreate` takes line items keyed by **variant** gid:

```graphql
lines: [{ merchandiseId: "gid://shopify/ProductVariant/42", quantity: 2 }]
```

Good news: **the variant id is already in the cart.** `toCartItem()` in `src/lib/product.ts`
encodes it into the cart key:

```ts
id: v ? `${product.id}#${v.id}` : product.id,
```

So a line reads `product:1#42`, and `42` is the numeric variant id. Nothing is missing, and the
existing design is sound — that helper is the single construction point for cart ids (the module
says so explicitly, and #63 fixed the wishlist by routing it through the same helper).

## The problem

The variant id is *recoverable* but not *addressable*. Getting it out means splitting a composite
key on `#` at the point of purchase — re-parsing a string that a helper deliberately owns the
construction of. That's fragile, and it puts knowledge of the key format in a second place.

## Proposed change

Add an explicit field to `CartLine` in `src/lib/product.ts`:

```ts
export interface CartLine {
  id: string;
  /** Numeric Shopify variant id, or null for lines with no Shopify variant
      (bundles, placeholder products). Null lines are not purchasable. */
  variantId: string | null;
  name: string;
  category: string;
  price: number;
  image: string;
}
```

…and set it in `toCartItem()` alongside the existing key. `CartProduct` in
`src/context/CartContext.tsx` extends the same shape, so it flows through.

The composite key stays exactly as it is — it does real work (two sizes of one product are
distinct rows in the bag) and changing it would churn `#20` and `#63`. This adds a field; it
does not replace the key.

## The lines that have no variant

This is the part that needs a decision, and it should be made here rather than discovered
during the checkout build:

- **Bundles** — `BundleCard.tsx` builds `bundle:${bundle.id}` directly, bypassing `toCartItem()`
  entirely. Bundles are hardcoded in `src/lib/products.ts` with no Shopify equivalent (#65), so
  they have no variant and **cannot be checked out**. Today a customer can add one to the bag.
- **No-variant products** — `toCartItem()` falls back to a bare `product.id` when
  `defaultVariant()` returns nothing. This is the placeholder / unconfigured-store path. Already
  guarded by the `placeholder` flag, but checkout should reject explicitly rather than rely on it.

Either is fine as a resolution; both need to be deliberate:
- resolve #65 by modelling bundles as real Shopify products (that issue's own first option, and
  it makes the hardcoded module disappear), **or**
- exclude `variantId === null` lines from the bag until #65 lands.

What must not happen is shipping a bag that accepts a bundle and then fails at checkout.

## Migration

Carts already in `localStorage` will have lines without `variantId`. `CartContext` has a
precedent for exactly this — it drops legacy numeric ids from before #20 on load. Extend that
same filter rather than writing a migration.

## Acceptance criteria

- [ ] `CartLine.variantId` exists and is populated by `toCartItem()`
- [ ] `BundleCard` either produces a real variant id or its lines are marked unbuyable
- [ ] Lines restored from `localStorage` without `variantId` are handled, not crashed on
- [ ] A decision is recorded on bundles (model in Shopify, or exclude from bag)
- [ ] `npm run lint` / `typecheck` / `build` pass

## Related

- Blocks #31 (payment gateway)
- Completes the last box on #4 (`[ ] Checkout handoff`)
- Interacts with #65 (bundles hardcoded) and #9 (bundle experience)
- Context: `docs/launch-readiness.md`

Oreginal:
Closes #92. Prerequisite for #31 (hosted checkout handoff); completes the last box on #4.

## Why

`cartCreate` needs `merchandiseId: "gid://shopify/ProductVariant/<id>"` per line. That id was
**already in the cart** — `toCartItem()` encodes it into the key as `product:1#42` — but only
*recoverably*, not *addressably*. Reaching it meant splitting a composite key at the point of
purchase, putting knowledge of the key format in a second place. Nothing in the codebase parses a
cart id apart today, and #31 shouldn't be the first thing to.

## What changed

**`variantId: string | null`, required on both `CartLine` and `CartProduct`** (they aren't linked
by a type alias). Required rather than optional is the point of the change — the compiler then
names every producer. It found exactly two, and both needed to change anyway:

- **`BundleCard`** — the one `addItem` call site that bypassed `toCartItem()`, building
  `bundle:<slug>` from an inline literal.
- **`WishlistSidebar.toBagLine`** — the `cartLine ?? product` fallback, already unreachable at
  runtime because `WishlistContext` drops entries without a `cartLine` (#63).

Numeric id, not a gid: nothing else in the client model holds one, and #31 rebuilds it at
mutation time.

**Bundles no longer enter the bag.** Hardcoded with no Shopify equivalent (#65), so a bundle line
could never produce a `merchandiseId` — today a shopper can fill a bag with something the store
cannot sell. The card now links to the catalogue. A real purchase path returns with #65.

**Legacy cart entries are dropped on restore**, mirroring the wishlist's `cartLine` handling
(#63): a stored line can't be repaired because the variant list isn't in it. Safe here because
`clearLocalUserState()` already wipes the key on sign-out and nothing is in production yet.

## Deliberately not in scope

**Stock is untouched — a sold-out variant still produces a line.** That guard belongs to #91,
which says it needs a UX decision (and possibly a refetch) first. This PR handles only lines that
are *structurally* unbuyable.

## Verification

`npm run lint` ✅ · `npm run typecheck` ✅ · `npm run build` ✅

There is no test suite in the repo (CI's Test step is `npm test --if-present`, a documented
no-op), so `toCartItem` was exercised directly against all 14 development-catalogue products:

- every product yields a non-null `variantId`
- `variantId` always equals the `#` suffix in the key, for every variant of every product
- multi-variant (`9001`): two sizes → two distinct rows, correct per-variant pricing
- cheapest-variant-sold-out (`9005`): `defaultVariant()` skips `90051`, line carries `90052`
- single-variant (`9003`): `variantId` set, no `— <label>` suffix on the name
- fully sold out (`9009`): still yields a variantId, confirming #92 doesn't guard stock
- variant-less "Coming Soon" tile: `variantId === null`, key has no `#`
- restore filter: drops pre-#92 lines and pre-#20 numeric ids, keeps an explicit `null`
  (`null !== undefined` — a known-absent variant is well-formed; #31 rejects it later)

Bundle cards confirmed rendering as `<a href="/store">` against a running dev server.

> **Note for the reviewer:** `next dev` appends a `nextjs-agent-rules` block to `CLAUDE.md`. I
> stripped it so this PR stays focused — it regenerates on the next `next dev` run. Worth
> committing separately if you want a clean tree.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -115,7 +115,10 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
 ## Conventions
 
 - **Cart/wishlist item ids are namespaced strings** — `product:<n>` (the numeric half of the
-  Shopify product gid) and `bundle:<slug>` (e.g. `bundle:glow`). Never bare numbers.
+  Shopify product gid). Never bare numbers. **Bundles no longer enter the cart** (#92): with no
+  Shopify variant they have no `merchandiseId` and could never be checked out, so `BundleCard`
+  links to the catalogue instead of adding to the bag. `bundle:<slug>` is therefore no longer a
+  cart key — restoring one depends on modelling bundles as Shopify products (#65).
   **The cart owns the variant suffix; the wishlist does not** (#63): cart lines append the
   Shopify variant id (`product:1#4567`) so two sizes are distinct rows, while a wishlist entry
   is a *product* and stays `product:1` — `isWishlisted(product.id)` is what fills the heart.
@@ -128,6 +131,13 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   the variant list isn't stored.
   Build every cart id with `toCartItem()` from `src/lib/product.ts` — it is the only place one
   is constructed. See `CartContext`.
+  **A cart line also carries `variantId`** (#92) — the same numeric variant id the `#` suffix
+  encodes, exposed as a field so the checkout handoff never splits the key apart. It is
+  `string | null` and **required**, so the compiler names every producer; `null` means the line
+  has no Shopify variant and cannot be bought, which #31 must reject rather than assume. Store
+  the numeric id, not a gid — `gid://shopify/ProductVariant/<id>` is rebuilt at mutation time.
+  Cart entries persisted before the field existed are dropped on restore, same reasoning as
+  `cartLine` above.
 - **Catalog data** (#4): products are fetched in `src/lib/shopify.ts` — `getProducts`,
   `getFeaturedProducts`, `getProductBySlug`; the `defaultVariant` / `isSoldOut` /
   `toCartItem` helpers live alongside the types in `src/lib/product.ts`. Fetches are tagged
@@ -135,7 +145,8 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   the tag on demand via `POST /api/revalidate`.
   With no credentials the store degrades to non-purchasable "Coming Soon" placeholders;
   a *configured* store that errors renders an empty grid rather than fake products.
-  Bundles are still hardcoded in `src/lib/products.ts` — no Shopify equivalent yet.
+  Bundles are still hardcoded in `src/lib/products.ts` — no Shopify equivalent yet, and
+  display-only since #92.
 - **Webhook auth** (#4): `/api/revalidate` verifies Shopify's `X-Shopify-Hmac-Sha256`
   (base64 HMAC-SHA256 over the raw body) with `timingSafeEqual`. Needs
   `SHOPIFY_WEBHOOK_SECRET` — the value Shopify shows under Settings → Notifications →
--- a/src/app/store/StoreContent.tsx
+++ b/src/app/store/StoreContent.tsx
@@ -63,6 +63,20 @@ export default function StoreContent({ products }: { products: ShopifyProduct[] 
 
   const clearCategory = useCallback(() => setActiveCategory(ALL_CATEGORIES), []);
 
+  /**
+   * On this page "Shop Bundle" would otherwise link to `/store` — the page the
+   * shopper is already on. Send them to the collection instead, reusing the same
+   * measured scroll `StoreHero` uses. The card keeps `#collection` as its href so
+   * the control still works without JS.
+   */
+  const handleBundleShop = useCallback(
+    (e: React.MouseEvent<HTMLAnchorElement>) => {
+      e.preventDefault();
+      scrollToCollection();
+    },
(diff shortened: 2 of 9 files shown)


Oreginal:
## Code review findings addressed — `30e21be`

Both findings confirmed as real and fixed.

### 1. Stale wishlist entries put an `undefined` variantId in the bag (HIGH)

Correct catch, and a genuine gap in my original plan: I guarded the cart's own storage key but not the **other** store that persists a `CartLine`.

An entry saved after #63 but before #92 has a `cartLine` — it just has no `variantId` on it. `cartLine != null` let it through, and `toBagLine` returned it verbatim, since the `??` fallback only covers a *missing* line, not a stale-shaped one. Reproduced end to end before fixing:

```
1. survives wishlist restore?   true
2. variantId is:                undefined
3. after localStorage round-trip: {"id":"product:9001#90011", ... }   <- key omitted
4. survives cart restore?       false   <- row silently vanishes
```

The wishlist filter now probes the captured line's field rather than the line's presence. I also took the suggestion on the cart filter and made it a positive check (`typeof === "string" || === null`), so a number from a hand-edited store is treated as legacy rather than trusted.

### 2. "Shop Bundle" was a dead control on /store (LOW/MEDIUM)

Also correct — `BundleCard` renders from both `sections/Bundles.tsx` and `store/StoreContent.tsx`, so the hardcoded `/store` link was a self-link on the latter, with an aria-label promising something the click didn't deliver. A regression I introduced.

The destination is now a prop. Home passes `/store`; the store page passes `#collection` plus a handler that scrolls to the collection using the same measured scroll `StoreHero` uses — the codebase deliberately measures the navbar instead of using `scroll-margin-top`, so a bare anchor would have landed underneath it. The href stays as the no-JS fallback.

### Verification

`lint` · `typecheck` · `build` pass. Regression checks: the finding-1 scenario now drops at wishlist restore; a post-#92 entry still round-trips through move-to-bag with its variantId intact; pre-#63 entries still drop unchanged; the hardened cart filter rejects numeric and absent variantIds while keeping an explicit `null`. The original 22 catalogue checks still pass.

Issue #67, opened 2026-08-12
Credential hygiene follow-up for the Shopify webhook migration


Oreginal:
Two Shopify secrets should be treated as exposed. Neither is urgent today; both become serious the
moment a higher-stakes webhook consumer exists.

## 1. The current webhook signing secret

During verification of #62 the store's webhook signing secret was pasted into a terminal command
line. It was written to the shell history file and to the session transcript on disk. It has not been
committed to the repository and has not left the machine, but it is no longer a value only Shopify
and Netlify know.

**Current blast radius is small.** The only consumer today is `POST /api/revalidate`, whose entire
effect is invalidating the `products` cache tag. Someone holding the secret can force redundant
Shopify re-fetches. That is a nuisance, not a breach — no data is readable and nothing is writable.

**Why it still matters:** this key signs *every* webhook the store sends, not just this one. The
moment a consumer is added for orders, customers, or fulfilment, the same secret authenticates those
payloads too, and a forged delivery stops being harmless.

- [ ] Check whether Shopify admin → Settings → Notifications → Webhooks offers a regenerate option
      for the signing secret.
- [ ] If it does: regenerate, then update `SHOPIFY_WEBHOOK_SECRET` in Netlify for **all** contexts and
      redeploy. Env vars are baked in at build time, so an existing deploy keeps the old value until
      rebuilt.
- [ ] If it does not: leave it, and treat rotating it as a prerequisite for wiring up any webhook
      beyond cache revalidation.
- [ ] Scrub the local shell history:
      ```
      $p="$env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine\ConsoleHost_history.txt"
      (Get-Content $p) | Where-Object { $_ -notmatch 'SECRET=' } | Set-Content $p -Encoding utf8
      ```

## 2. The old `SHOPIFY_REVALIDATE_SECRET`

Before #62, `/api/revalidate` authenticated via `?secret=<value>` in the query string. Query strings
are logged by CDNs, proxies and web servers as a matter of course, and leak through `Referer`
headers. That value should be assumed to sit in access logs already.

**Deleting the environment variable does not remove it from historical logs.** Retire the value
rather than reuse it anywhere. Deletion from Netlify is tracked in the go-live issue.

Related: #62, #36.

Issue #66, opened 2026-08-12
Shopify production env + webhook registration before main -> master


Oreginal:
Blocks going live. Everything here is Shopify/Netlify configuration — no code changes.

PR #62 wired the storefront to Shopify and replaced the old query-string webhook auth with HMAC
verification. All of it is **verified working on the deploy preview**, but the preview and production
are separate Netlify contexts, so none of that configuration carries over automatically.

**Do not follow the checklist in #36** — it predates #62 and tells you to register webhooks as
`https://<url>/api/revalidate?secret=<value>`. That endpoint no longer reads a query parameter. A
webhook registered that way returns **401** on every call, and Shopify reports it as a delivery
failure with no other symptom on the site.

## Before merging `main` → `master`

- [ ] **`SHOPIFY_WEBHOOK_SECRET` is set for the Production context** in Netlify → Site configuration →
      Environment variables. It is currently set and verified for Deploy previews; Netlify scopes
      variables per context, so Production may still be blank. No `NEXT_PUBLIC_` prefix — this value
      must never reach the browser.
- [ ] **`NEXT_PUBLIC_SHOPIFY_STORE_DOMAIN` and `NEXT_PUBLIC_SHOPIFY_STOREFRONT_TOKEN` are set for the
      Production _build_ environment.** Without them the build prerenders "Coming Soon" placeholders
      and `generateStaticParams` emits zero paths — the store ships empty, with no build error.
- [ ] **Delete `SHOPIFY_REVALIDATE_SECRET` from Netlify.** Nothing reads it after #62. See #67 for why
      the old value should be treated as compromised rather than reused.

## After the production deploy is live

- [ ] **Register the webhooks** in Shopify admin → Settings → Notifications → Webhooks:
      - Topics: `products/create`, `products/update`, `products/delete`
      - Format: **JSON**
      - API version: **2026-04** — must match `SHOPIFY_API_VERSION` in `src/lib/shopify.ts`. A
        mismatch changes the payload shape Shopify signs.
      - URL: `https://<live-domain>/api/revalidate` — **no query string.**
- [ ] **Verify with a real change**: edit a product price in Shopify, then load the live `/store`. The
      new price must appear on the **next** request. If it takes two requests, or up to an hour, the
      webhook isn't reaching the site — check Shopify's delivery log first.

## How to tell what's wrong from the status code

The endpoint distinguishes these deliberately, so Shopify's delivery log is diagnostic:

| Status | Meaning |
[...]

Issue #18, opened 2026-05-30
Production Deployment


Oreginal:
## Description\nPrepare and deploy production version.\n\n## Acceptance Criteria\n- [ ] Environment variables configured\n- [ ] Shopify connected\n- [ ] Domain connected\n- [ ] SSL active

PR #94, opened 2026-08-25
feat: hand off to Shopify hosted checkout (#31)

### LINKED ISSUE #31: Integrate the payment gateway with Shopify (hosted checkout handoff)

Oreginal:
Closes #31. Builds on #92 (`variantId` on the cart line). Completes the last box on #4.

`/checkout` ended at a dashed "Secure payment coming soon" pill. The bag now becomes a Shopify
cart and the shopper is sent to the returned `checkoutUrl` — Shopify owns payment, PCI scope,
tax, shipping rates and the order record. **PayFast is configured in the Shopify admin; no
gateway code is in this PR, and none belongs here.**

## ⚠️ The happy path is NOT verified — please read

The local Storefront credentials are rejected by Shopify. Even a trivial read fails:

```
POST https://wi9fzy-ha.myshopify.com/api/2026-04/graphql.json   { shop { name } }
→ HTTP 401
```

So no cart can be created from this machine, and **I could not confirm that a shopper actually
reaches Shopify's checkout.** Everything up to the Shopify call is verified; the call itself
isn't. This looks like #66's territory (production Storefront credentials) rather than a code
problem — the domain matches the connected store, and the token is the right shape but doesn't
authenticate.

Worth knowing: `NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS=true` **masks this completely**. The site
browses perfectly on the dev catalogue while the token is dead, so nothing else surfaces it.

Before merging, someone with working credentials should confirm: Proceed → Shopify checkout,
right line items and quantities, totals matching the summary, email pre-filled.

## Three things the docs settle

Each is a quiet way to get this wrong, so they're worth calling out:

1. **The Cart API reports failures in `userErrors`, inside `data`, with a 200.** `shopifyFetch`'s
   `json.errors` check never sees them — a rejected cart would look like a successful one.
   `createCart` checks them explicitly.
2. **Mutations must not reuse `shopifyFetch`.** It sets `next: { revalidate: 3600, tags: ["products"] }`,
   which would cache a cart creation for an hour under the products tag — two shoppers could be
   handed the same checkout. `storefrontMutate` is its `cache: "no-store"` sibling.
3. **A cart id embeds a secret** (`gid://shopify/Cart/<token>?key=<secret>`); Shopify says treat it
   like a password. The mutation selects `cart { checkoutUrl }` and nothing else, so the id can't
   leak through this route by accident.

## Why `/api/checkout`

`proxy.ts` gates every path starting with `/checkout` and its matcher excludes `/api`. A handler
under `/checkout/...` — or a Server Action, which POSTs to the page's own URL — would answer a
stale-cookie POST with a **307 to `/login`**, handing `fetch` an HTML body. Living under `/api`
avoids that entirely.

**It is unauthenticated, deliberately.** It grants nothing a visitor doesn't already have —
Shopify's cart permalinks (`/cart/<variantId>:<qty>`) build a checkout URL with no API at all. It
reads no user data, takes no payment, and returns only a URL Shopify hands out freely. Verifying a
Firebase ID token would need `firebase-admin`, which isn't in the tree, and under hosted checkout
there's no order endpoint of ours left to protect (#57). What it does instead is validate input
tightly so it can't become a general-purpose Storefront proxy. **If you disagree with that call,
this is the thing to push back on.**

The `email` is a checkout **prefill, not an identity claim** — it comes from the browser, and #57
is explicit that client-sent values aren't a trust boundary.

## Other decisions

- **The bag is kept on handoff.** An abandoned payment or a back-button press must not cost the
  shopper their bag. Clearing on confirmation needs an `orders/create` webhook (#67).
- **`checkoutUrl` is origin-checked** before being returned. `redirect.ts` is the boundary for
  *internal* destinations and rejects every absolute URL by design, so this is its outbound
  counterpart — the value becomes a top-level navigation, so an unexpected host must not be followed.
- **`pending` is not reset on success** — the browser is still navigating away, and resetting
  flashes the button back to idle.
- No new env vars. Noted in `.env.example`, since a payment step is exactly where you'd expect a
  secret to be needed.

## Verified

`npm run lint` ✅ · `npm run typecheck` ✅ · `npm run build` ✅ (`/api/checkout` registers)

Live endpoint, against a running server:

| Request | Result |
| --- | --- |
| empty bag | `400 {"message":"Your bag is empty."}` |
| `variantId: null` (#92 unbuyable) | `400 {"message":"That bag contains an item that can't be purchased."}` |
| non-numeric id | `400` |
| quantity `0` | `400` |
| malformed body | `400` |
| dev-catalogue variant | `502` + friendly message, bag intact |

URL guard: accepts the shop domain and `*.shopify.com`; rejects `http:`, foreign hosts,
`evil-shopify.com`, `shopify.com.evil.com`, `javascript:` and garbage. Input validation covers
fractional/negative quantities and the 50-line cap.

Not exercised: the rendered button (the Chrome extension isn't connected, and `/checkout` is
auth-gated). It's confirmed present in the built client chunk.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,6 +29,12 @@ NEXT_PUBLIC_SHOPIFY_STOREFRONT_TOKEN=
 # Production must be `false` or unset.
 NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS=false
 
+# CHECKOUT (#31) NEEDS NO NEW VARIABLES.
+# The hosted-checkout handoff reuses the two Storefront values above — the cart is
+# created with the same public Storefront token, and Shopify runs payment itself.
+# The payment gateway (PayFast) is configured in the Shopify admin, not here, so
+# no gateway key, merchant id or passphrase belongs in this file.
+
 # Shopify webhook signing secret, used to verify the HMAC on POST /api/revalidate.
 # Shopify admin → Settings → Notifications → Webhooks, "Your webhooks are signed
 # with this secret" at the bottom of the page. This is NOT a value you invent.
--- a/.env.example
+++ b/.env.example
@@ -29,8 +29,15 @@ NEXT_PUBLIC_SHOPIFY_STOREFRONT_TOKEN=
 # Production must be `false` or unset.
 NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS=false
 
-# CHECKOUT (#31) NEEDS NO NEW VARIABLES.
-# The hosted-checkout handoff reuses the two Storefront values above — the cart is
+# Shopify's `checkoutUrl` comes back on the shop's PRIMARY domain, which is not
+# necessarily the API domain above. Once a custom domain is connected (#18), set
+# this to it — e.g. https://shop.heybeautiful.co.za — or the checkout handoff
+# will reject its own URL and every shopper gets a generic failure.
+# Unset is correct while the store is still on *.myshopify.com.
+NEXT_PUBLIC_SHOPIFY_CHECKOUT_DOMAIN=
+
+# CHECKOUT (#31) NEEDS NO OTHER NEW VARIABLES.
+# The hosted-checkout handoff reuses the Storefront values above — the cart is
 # created with the same public Storefront token, and Shopify runs payment itself.
 # The payment gateway (PayFast) is configured in the Shopify admin, not here, so
 # no gateway key, merchant id or passphrase belongs in this file.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -147,6 +147,25 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   a *configured* store that errors renders an empty grid rather than fake products.
   Bundles are still hardcoded in `src/lib/products.ts` — no Shopify equivalent yet, and
   display-only since #92.
+- **Checkout handoff** (#31): the bag becomes a Shopify cart via `createCart()` in
+  `src/lib/shopify.ts`, and the shopper is sent to the returned `checkoutUrl`. Shopify owns
+  payment, PCI scope, tax, shipping and the order record; the gateway (PayFast) is configured
+  in the Shopify admin, **never in this repo**.
+  `POST /api/checkout` is the only mutation path. It lives under `/api` on purpose: `proxy.ts`
+  gates everything starting with `/checkout` and excludes `/api`, so a handler under
+  `/checkout/...` — or a Server Action, which posts to the page's own URL — would be answered
+  with a 307 to `/login` on a stale cookie and hand `fetch` an HTML body. It is unauthenticated
+  because it grants nothing a visitor lacks (Shopify cart permalinks do the same with no API),
+  so it validates input tightly instead: numeric variant ids, integer quantities, capped line
+  count. The `email` it receives is a checkout **prefill, not an identity claim** (#57).
+  **Never select `cart { id }`** — a cart id embeds a secret `?key=` that Shopify says to treat
+  like a password. Only `checkoutUrl` is ever requested or returned.
+  Mutations use `storefrontMutate()`, **not** `shopifyFetch()`: the latter caches for an hour
+  under the `products` tag, which would hand two shoppers the same checkout. `userErrors` must
+  be checked explicitly — the Cart API reports business failures inside `data` with a 200, so
+  the `json.errors` check never sees them.
+  The bag is **kept** on handoff, not cleared: an abandoned payment or a back-button press must
+  not cost the shopper their bag. Clearing on confirmation needs an `orders/create` webhook (#67).
 - **Webhook auth** (#4): `/api/revalidate` verifies Shopify's `X-Shopify-Hmac-Sha256`
   (base64 HMAC-SHA256 over the raw body) with `timingSafeEqual`. Needs
   `SHOPIFY_WEBHOOK_SECRET` — the value Shopify shows under Settings → Notifications →
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,6 +166,18 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   the `json.errors` check never sees them.
   The bag is **kept** on handoff, not cleared: an abandoned payment or a back-button press must
   not cost the shopper their bag. Clearing on confirmation needs an `orders/create` webhook (#67).
+  Because `pending` is deliberately left set while the browser navigates away, `CheckoutContent`
+  resets it on `pageshow` with `event.persisted` — a bfcache restore (Back from Shopify) otherwise
+  returns the shopper to a permanently disabled button.
+  `checkoutUrl` comes back on the shop's **primary** domain, which is not necessarily the API
+  domain — set `NEXT_PUBLIC_SHOPIFY_CHECKOUT_DOMAIN` once a custom domain is connected (#18), or
+  the guard rejects Shopify's own URL. **Never log a checkout URL**: it carries the cart's secret
+  `?key=`. Log the host.
+  Server-side Storefront calls forward `Shopify-Storefront-Buyer-IP`. Shopify meters a public
+  token per client IP, so without it every shopper's cart shares one bucket with the catalogue
+  reads and a flood could empty the store's grid.
+  Bag caps (`MAX_CART_QUANTITY`, `MAX_CART_LINES`) live in `@/lib/constants` so the cart UI and the
+  endpoint enforce the same ceiling — a server-only limit is a dead end the shopper can't diagnose.
 - **Webhook auth** (#4): `/api/revalidate` verifies Shopify's `X-Shopify-Hmac-Sha256`
   (base64 HMAC-SHA256 over the raw body) with `timingSafeEqual`. Needs
   `SHOPIFY_WEBHOOK_SECRET` — the value Shopify shows under Settings → Notifications →
--- a/src/app/api/checkout/route.ts
+++ b/src/app/api/checkout/route.ts
@@ -1,0 +1,124 @@
+// Turns the shopper's bag into a Shopify cart and returns its checkout URL (#31).
+// The browser then navigates there and Shopify runs the rest of the purchase.
+//
+// WHY /api AND NOT A SERVER ACTION OR /checkout/...:
+// `proxy.ts` gates every path starting with `/checkout` and its matcher excludes
+// `/api`. A handler under `/checkout/...` — and a Server Action, which POSTs to
+// the page's own URL — would therefore be answered with a 307 to `/login` the
+// moment the session-hint cookie is stale, and `fetch` would follow it and get
+// HTML back. Living under `/api` avoids that whole failure mode.
+//
+// WHY THIS ENDPOINT IS UNAUTHENTICATED:
+// It grants no capability a visitor doesn't already have. Shopify's own cart
+// permalinks (`/cart/<variantId>:<qty>`) let anyone build a checkout URL with no
+// API at all. This reads no user data, takes no payment, and returns only a URL
+// Shopify is happy to hand out. Verifying a Firebase ID token would need
+// `firebase-admin`, which this repo doesn't carry — that's #57's ground, and
+// under hosted checkout there is no order endpoint of ours left to protect.
+// What it MUST do instead is validate its input tightly, so it can't be turned
+// into a general-purpose proxy for arbitrary Storefront mutations.
+
+import { NextResponse, type NextRequest } from "next/server";
+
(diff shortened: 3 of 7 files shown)


Oreginal:
## Code review findings addressed — `60ad34a`

All five were real. Two of them would have broken checkout in production.

### 1. HIGH — the URL guard rejected Shopify's own `checkoutUrl`

The sharpest catch. `"shop.myshopify.com".endsWith(".shopify.com")` is **false**, so that arm never covered `*.myshopify.com` and the whole allowlist rested on an exact match against the configured API domain.

But Shopify returns `checkoutUrl` on the shop's **primary** domain, which isn't necessarily the API domain. So the day a custom domain is connected — which is literally a checkbox in #18 — every shopper would have got a generic 502, with the only diagnostic a server-side `console.error`.

The guard now accepts Shopify-owned hosts explicitly (including `*.myshopify.com`) plus an optional `NEXT_PUBLIC_SHOPIFY_CHECKOUT_DOMAIN`. It also logs when the store-domain env is set without its scheme, instead of silently narrowing the allowlist as the old IIFE did.

### 2. MEDIUM — Back from Shopify left the button permanently disabled

`pending` is deliberately not reset on success because the browser is still navigating — but a bfcache restore brings React state back with it. The shopper returned to a button stuck on "Taking you to checkout…" and couldn't retry short of a hard reload.

The bag survived and checkout didn't, which defeats the point of keeping the bag. Reset on `pageshow` when `event.persisted`.

### 3. MEDIUM — cart creation burned the server's rate-limit bucket

Correct, and the argument I used to justify the unauthenticated endpoint genuinely doesn't carry here. Shopify meters a public token **per client IP**; routed through our server, every cart shared one bucket with the catalogue reads in `shopifyFetch` — so a flood could throttle `getProducts()` and empty the grid site-wide.

Now forwards `Shopify-Storefront-Buyer-IP`, which is Shopify's documented remedy for server-side Storefront calls. A true per-IP throttle would need infrastructure this repo doesn't have; this is the prescribed mitigation.

### 4. MEDIUM — the rejection log wrote the cart's secret

A checkout URL carries `?key=<secret>` — the same value the mutation avoids selecting `cart { id }` to protect. Combined with finding 1 it would have fired on **every** checkout, writing a live cart capability per shopper into Netlify's function logs. Logs the host now.

### 5. LOW — the caps had no UI counterpart

Moved `MAX_CART_QUANTITY` / `MAX_CART_LINES` into `@/lib/constants`, clamped `updateQuantity` to the same ceiling, and both messages now name the limit.

## Verification

`lint` · `typecheck` · `build` pass.

Guard: accepts any `*.myshopify.com` and a configured primary domain; still rejects `http:`, foreign hosts, `evil-shopify.com`, `shopify.com.evil.com`, `notmyshopify.com.evil.io` and `javascript:`. Verified it doesn't go soft — the added arms are host-suffix checks, not substring matches.

Endpoint: `{"message":"Each item is limited to 100. Please reduce the quantity."}` and `{"message":"A bag can hold at most 50 different items."}`. Server log confirmed free of raw URLs and secrets.

**Still unverified: the happy path.** The Storefront token continues to return 401, so no cart can be created here. Unchanged from the original PR body.

Oreginal:
## Verified end to end against the live Shopify store

The Storefront credentials that previously 401'd are fixed (wrong token type — the Headless
channel's **private** token was being sent in the **public** token's header). With working
credentials I've now exercised this PR against the real store rather than the development
catalogue, driving our own `POST /api/checkout` rather than raw GraphQL.

### Acceptance criteria (#31)

| Criterion | Result |
| --- | --- |
| "Proceed to payment" creates a Shopify cart and returns `checkoutUrl` | ✅ 200 with a real hosted-checkout URL |
| Line items carry the correct variant and quantity | ✅ 2 × one product, 1 × another, as sent |
| **Shopify's totals match the summary shown** | ✅ **R 410,99** = 2 × R 100 + R 210,99, ZAR |
| Response contains only `checkoutUrl`, no `cart.id` | ✅ the secret `?key=` is never selected |
| `userErrors` and network failures surface to the customer | ✅ unknown variant → 502 + friendly message, no crash, no blank page |
| Unbuyable lines cannot reach this path | ✅ `variantId: null`, non-numeric, and a full gid all rejected with 400 |
| Empty bag cannot trigger a handoff | ✅ 400 "Your bag is empty." |
| `npm run lint` / `typecheck` / `build` | ✅ all pass |

### Input validation

Every case below was sent as a hand-rolled POST — the UI can't produce most of them:

| Input | Response |
| --- | --- |
| `lines: []` | 400 — bag is empty |
| `variantId: null` | 400 — not purchasable |
| non-numeric `variantId` | 400 — not purchasable |
| full gid instead of numeric id | 400 — not purchasable |
| `quantity: 0` / `-5` / `1.5` | 400 — invalid quantity |
| `quantity: 100` (at `MAX_CART_QUANTITY`) | ✅ 200 — boundary passes |
| `quantity: 101` | 400 — over the cap |
| 51 lines (`MAX_CART_LINES` is 50) | 400 — too many items |
| missing `lines` / `lines` not an array | 400 — malformed |

### One fix came out of this

`ed444fa` — quantity `0`, `-5` and `1.5` were all rejected with *"Each item is limited to 100"*,
which names the wrong problem. Malformed quantities now get their own message. Only reachable by
a hand-rolled POST, since the cart UI clamps, but the endpoint should say what it actually refused.

### Side effect worth noting

With real credentials the build now prerenders the product detail routes as SSG rather than
falling back — it could not do that while the token was dead.

### What this does *not* prove

**No payment has been taken.** Shopify's hosted checkout was not driven to completion, because no
payment gateway is installed on the store yet and no shipping rates are configured. Treat this as
code-complete and integration-verified, not revenue-proven — a real transaction placed and
refunded is still outstanding, and that's admin work rather than anything in this diff.

Issue #14, opened 2026-05-30
SEO Foundation


Oreginal:
## Description\nPrepare website for search engines.\n\n## Acceptance Criteria\n- [ ] Metadata\n- [ ] Open Graph\n- [ ] Structured data\n- [ ] Sitemap\n- [ ] Robots.txt\n- [ ] llm.txt

Issue #68, opened 2026-08-12
Development product catalogue behind NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS


Oreginal:
The Shopify catalogue isn't populated yet — `/store` currently renders **three** products. That
blocks storefront design work: no layout can be judged against three cards, and the store redesign
shouldn't wait on merchandising.

Add a development catalogue behind an environment switch so the frontend can be finished now and
transition to real products later without another redesign.

## Behaviour

| `NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS` | Source |
| --- | --- |
| `true` | local development catalogue |
| `false` | real Shopify catalogue |
| unset / anything else | real Shopify catalogue (safe default) |

Only the exact string `true` enables it. `NEXT_PUBLIC_*` values are inlined at **build** time, so
changing this on Netlify requires a new build — it does not affect a running app.

## Scope

- New `src/lib/placeholder-products.ts` exporting a catalogue that satisfies the **existing**
  `ShopifyProduct` contract. No parallel product interface, no changes to the product model.
- Switch inside `getProducts()` and `getProductBySlug()` so home, store and `/store/[slug]` all
  honour it. `getFeaturedProducts()` and `generateStaticParams()` inherit it.
- A deterministic `getShowcaseProducts()` selector — `featured`-tagged products first, then a stable
  top-up in catalogue order. Never random.
- Document the flag in `.env.example` and `CLAUDE.md`.

## The catalogue must exercise the components, not just fill space

12–16 products spanning Supplements / Protein / Beauty / Wellness / Recovery, and including:
multiple variants and single variants; discounted and full price; a sold-out product; **a product
whose cheapest variant is sold out while a pricier one is discounted** (the case #62's per-variant
`compareAtPrice` fix targets, still unreproducible against the real store); with and without
ratings; with and without badges; two images per product for the card hover swap; long and short
names.

Every product needs a stable `slug` that resolves through the existing `/store/[slug]` route.

## Naming hazard

`ShopifyProduct.placeholder?: boolean` already means *"non-purchasable Coming Soon tile"* and is used
by `ShopifyProductCard` to strip every purchase control. This flag means the opposite — realistic
data that must behave exactly like real products. **Dev-catalogue products must never set
`placeholder: true`,** and the existing unconfigured-store fallback stays as it is.

## Acceptance

[...]

PR #97, opened 2026-08-26
feat: add sitemap.xml and robots.txt (#14)

Oreginal:
Covers the **Sitemap** and **Robots.txt** boxes on #14. The site previously had neither — no
`sitemap.ts`, no `robots.ts`, no `public/robots.txt` — so crawlers got no guidance at all, and
nothing stopped a Netlify deploy preview being indexed alongside production as duplicate content.

## What's here

| File | Purpose |
| --- | --- |
| `src/lib/site.ts` | **new** — resolves this deploy's public origin, and whether it may be indexed |
| `src/app/robots.ts` | **new** — context-aware rules + sitemap pointer |
| `src/app/sitemap.ts` | **new** — static routes + one URL per real product |
| `src/app/layout.tsx` | `metadataBase` now derives from the same resolver |
| `.env.example` | documents the optional `NEXT_PUBLIC_SITE_URL` |
| `CLAUDE.md` | conventions entry |

## Base URL

Resolved per deploy: `NEXT_PUBLIC_SITE_URL` → Netlify's `DEPLOY_PRIME_URL` / `URL` → localhost.
Ordering matters — on a preview build *both* Netlify vars are set, and `URL` points at production,
so preferring it would make every preview advertise production's URLs.

This works on Netlify today with no configuration. Attaching a real domain (#18) is then a single
env var plus a rebuild, with no code change.

**`src/lib/site.ts` is server-only and deliberately not in `@/lib/constants`.** `CONTEXT`, `URL` and
`DEPLOY_PRIME_URL` have no `NEXT_PUBLIC_` prefix, and `constants.ts` is imported by `CartContext`,
`WishlistContext` and the auth screens — putting it there would resolve every value to the localhost
fallback in the browser.

## Indexing

Production allows crawling; `deploy-preview` and `branch-deploy` get `Disallow: /` and **no**
`Sitemap:` line. `isIndexableDeploy()` fails **open** on an unrecognised `CONTEXT` — silently
de-indexing the real site is a failure nobody would notice until traffic vanished.

`robots.ts` keeps its own private-path list rather than importing the proxy's. They look like the
same list but answer different questions: `proxy.ts` is an enforced access boundary, robots.txt is a
request to well-behaved crawlers. Sharing one list would imply this file protects something.
(`/verify-email` is in ours but intentionally absent from the proxy's `AUTH_PAGES`.)

## Drive-by fix: `metadataBase`

It was hardcoded to `https://heybeautiful.com`, which **is not connected yet** (#18), so every
relative Open Graph image was resolving against a host that doesn't answer. Pointing it at the same
resolver fixes that and — the reason it belongs in this PR — stops the site carrying two disagreeing
base URLs.

## On `lastModified`

Omitted deliberately. `ShopifyProduct` has no date field of any kind, and the obvious substitute,
`new Date()`, would assert that every page changed every time the sitemap regenerates (hourly),
which teaches crawlers to discount the field. Adding a real `updatedAt` to the catalogue fragment is
a clean follow-up.

## Verification

`lint` · `typecheck` · `build` all pass. Beyond that, each case exercised against a running server:

| Case | Result |
| --- | --- |
| Default (local) | 2 static + 3 product URLs; robots lists all private paths |
| `CONTEXT=production` | Allows; sitemap + `Sitemap:` line use `URL` |
| `CONTEXT=deploy-preview` | `Disallow: /`, no `Sitemap:` line; sitemap uses `DEPLOY_PRIME_URL`, not `URL` |
| `NEXT_PUBLIC_SITE_URL` set | Overrides both; `og:image` picks up the same origin |
| Bad Shopify token | Degrades to the 2 static routes, HTTP 200, no crash |

Build output confirms `/sitemap.xml` carries the `1h` revalidate and the `products` tag, so the
existing Shopify webhook refreshes it, and its 3 product URLs match the 3 SSG routes exactly.

## Also in this diff

The `nextjs-agent-rules` block that `next dev` re-appends to `CLAUDE.md` on every run. Committing it
once stops it reappearing as an uncommitted change — which is what the block itself recommends.

## Still open on #14

Structured data (JSON-LD), `llm.txt`, per-product canonical + Open Graph, and a `favicon.ico`
(confirmed absent from `public/`).

Related: #96 adds the Search Console verification tag for the same issue.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,19 @@ NEXT_PUBLIC_FIREBASE_APP_ID=
 NEXT_PUBLIC_SHOPIFY_STORE_DOMAIN=
 NEXT_PUBLIC_SHOPIFY_STOREFRONT_TOKEN=
 
+# Public origin of THIS site, used for the sitemap, robots.txt and the absolute URLs
+# in Open Graph tags (`metadataBase`). Optional — leave it blank unless you need to
+# override the automatic value.
+#
+# Netlify supplies a working origin on its own: `URL` on production builds and
+# `DEPLOY_PRIME_URL` on deploy previews, which `src/lib/site.ts` reads. Set this only
+# once a custom domain is attached (#18) and you want it quoted instead — for example
+# https://heybeautiful.co.za  (include the scheme, no trailing slash).
+#
+# NEXT_PUBLIC_* values are inlined at BUILD time, so setting this in Netlify needs a
+# NEW BUILD; redeploying an existing one keeps the old value baked in.
+NEXT_PUBLIC_SITE_URL=
+
 # Development product catalogue (#68). Lets the storefront be built and reviewed
 # before the Shopify catalogue is populated.
 #
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -184,6 +184,19 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   `CheckoutContent` (the edge `proxy.ts` can't read Firebase's `emailVerified`), redirecting
   unverified users to `/verify-email?from=<dest>`. `/account` is advisory (soft reminder only).
   `AuthContext.reloadUser()` refreshes `emailVerified`; the verify page polls it to auto-continue.
+- **Site URL and indexing** (#14): `siteUrl()` in `src/lib/site.ts` is the **single** source of
+  the public origin — it feeds both `metadataBase` in `layout.tsx` and `sitemap.ts`, so the two
+  can never quote different hosts. It resolves `NEXT_PUBLIC_SITE_URL` → Netlify's
+  `DEPLOY_PRIME_URL`/`URL` → localhost; **server only**, since those Netlify vars carry no
+  `NEXT_PUBLIC_` prefix and would be `undefined` in a client bundle (which is why it isn't in
+  `@/lib/constants`, imported by the cart and auth contexts). `isIndexableDeploy()` blocks
+  crawling on deploy previews so they can't compete with production as duplicate content; it
+  fails **open** on an unrecognised `CONTEXT`, because silently de-indexing the real site is the
+  worse failure. `robots.ts` keeps its own private-path list rather than importing the proxy's —
+  the proxy is an access boundary, robots.txt is a request to well-behaved crawlers, and sharing
+  a list would imply robots.txt protects something. `sitemap.ts` must keep using the same
+  `filter((p) => p.slug)` predicate as `generateStaticParams`, or it will advertise URLs that
+  have no prerendered page.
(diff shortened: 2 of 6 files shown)


Oreginal:
## Verified on the live deploy preview

Beyond `lint` / `typecheck` / `build`, every branch of the new logic was exercised against a
running server — and the preview-blocking case was confirmed on **this PR's actual Netlify
preview**, not just simulated locally.

### On the live preview

`/robots.txt` returns exactly:

```
User-Agent: *
Disallow: /
```

No `Sitemap:` line, as intended — pointing crawlers at a sitemap you've just disallowed is
contradictory. **Before this PR the preview was fully crawlable** and competing with production as
duplicate content.

`/sitemap.xml` renders with the *preview* origin throughout, not production's. That's the case
ordering was written for: on a preview build Netlify sets both `URL` (production) and
`DEPLOY_PRIME_URL` (this deploy), so preferring `URL` would make every preview advertise
production's URLs.

### Local matrix

| Case | Result |
| --- | --- |
| Default | 2 static + 3 product URLs; robots lists every private path |
| `CONTEXT=production` | Allows; sitemap and `Sitemap:` line use `URL` |
| `CONTEXT=deploy-preview` | `Disallow: /`, no sitemap pointer; URLs use `DEPLOY_PRIME_URL` |
| `NEXT_PUBLIC_SITE_URL` set | Overrides both; `og:image` picks up the same origin |
| Bad Shopify token | Degrades to the 2 static routes, HTTP 200, no crash or build failure |

Build output confirms `/sitemap.xml` carries the `1h` revalidate and the `products` cache tag, so
the existing Shopify webhook refreshes it, and its product URLs match the SSG routes exactly.

---

## ⚠️ Unrelated finding this surfaced — worth acting on separately

The preview's sitemap lists **14 products, but the live Shopify store has 3.**

Handles including `morning-glow-ritual`, `iron-b12-support`, `radiance-marine-collagen` and
`rose-recovery-body-butter` come from `src/lib/placeholder-products.ts`, and the real `test-product`
is absent. That means **`NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS=true` is set on Netlify** — the
development catalogue (#68), not Shopify.

This is **not caused by this PR**, and the sitemap is behaving correctly: it faithfully mirrors
whatever catalogue is configured, exactly as `generateStaticParams` already does.

But it does raise the stakes on that flag. Previously a wrong value only meant fake tiles on a page
nobody was indexing. Once a sitemap exists, **a production build with the flag still on would
actively submit ~14 fabricated product URLs to Google** — pages that will 404 the moment the flag is
corrected. Since `NEXT_PUBLIC_*` is inlined at build time, fixing it requires a **new build**, not
just an env edit.

Suggested: confirm the value per deploy context in the Netlify UI before the `main → master`
promotion. Belongs with #66/#18 rather than here.

Issue #51, opened 2026-07-15
Product packaging artwork: label mismatches, printed typo, off-brand background


Oreginal:
Split out from #13 (Product Photography Review). #13 / PR #50 covered everything code can fix — sizing, compression, responsive delivery. **These defects need the artwork re-rendered and can't be fixed in code.**

Found while reviewing the mockups side by side on `/store`.

## 1. Packaging labels contradict the catalog

Four of the five products show packaging whose printed name doesn't match what we're selling:

| Catalog name (`src/lib/products.ts`) | Image | Packaging actually reads | |
|---|---|---|---|
| Glow Collagen Blend | `item-1` | **GLOW & GO PRE-WORKOUT / Litchi Pineapple** | ❌ wrong product entirely |
| Plant Protein Luxe | `item-2` | PLANT PROTEIN / Strawberry Cream | ✅ matches |
| Morning Glow Ritual | `item-3` | **PLANT PROTEIN / Chocolate Delight** | ❌ |
| Recovery Rose Blend | `item-4` | **PLANT PROTEIN / Chocolate Delight** | ❌ + duplicate of item-3's flavour |
| Strength & Radiance | `item-5` | **PLANT PROTEIN / Strawberry Cream** | ❌ + duplicate of item-2's flavour |

A customer on `/store` sees "Glow Collagen Blend" over a tub that says PRE-WORKOUT, and two adjacent cards both reading "Chocolate Delight". The five SKUs are really 2 flavours × 2 formats + 1 pre-workout.

## 2. "Protien" typo printed on the packaging

`item-2`, `item-3`, `item-4`, `item-5` all read:

> 22g Plant **Protien**

It's baked into the label artwork, so it ships on every product card, the detail gallery, cart, wishlist, and checkout. This is the most visible one — worth fixing first.

## 3. `item-1` background is off-brand

`item-1` sits on a **lavender/periwinkle** sweep; `item-2`–`item-5` are **blush pink**. The card container is `bg-[#f0ebe3]` (parchment) and the palette is rose-gold / ink / cream / parchment — the lavender matches nothing, and the navy lid is off-palette too. It's visibly the odd one out in the grid.

## 4. Please supply 1200×1500 native masters

Current sources are **1600×1066 landscape**; the site frames product imagery **4:5 portrait**. PR #50 crops them to 853×1066, which is the largest true 4:5 crop available — but that's a ceiling imposed by the source, not a choice.

[...]

PR #99, opened 2026-08-26
feat: add llms.txt and explicit AI crawler rules (#14)

Oreginal:
Closes the last box on **#14**.

## `/llms.txt`

Follows the [llmstxt.org](https://llmstxt.org/) convention — an H1, a blockquote summary, then H2
sections of annotated links — so an assistant can learn what this store sells in one cheap fetch
rather than crawling pages and stripping HTML.

**The filename is plural.** #14 writes it "llm.txt", but `llms.txt` is the name the convention
defines and the only one crawlers look for. Next's own docs ship at `/docs/llms.txt` for the same
reason.

It is a **Route Handler**, not a static file in `public/`, because the product list has to track the
Shopify catalogue. Reusing `getProducts()` means it inherits the hourly revalidate and the
`products` cache tag, so **the existing Shopify product webhook refreshes it** alongside the
sitemap. It reads nothing from the request, so it prerenders as static.

Product lines reuse `formatPrice` and the same `filter(p => p.slug)` predicate as `sitemap.ts` and
`generateStaticParams`, so the three can't drift.

## AI crawler rules

`robots.ts` now names the major AI crawlers and allows them. Hey Beautiful is a new brand with
nothing proprietary on the site, so being citable when someone asks an assistant for wellness
supplements is free distribution. Reversing it is a one-line change from `allow` to `disallow`.

**`PRIVATE_PATHS` is repeated for that group on purpose** — this is the part worth reviewing. Per
the robots.txt spec a crawler obeys only the *most specific* group matching its name and ignores `*`
entirely. Naming `GPTBot` without repeating the disallows would therefore **widen** its access to
`/account` and `/checkout` rather than leave it unchanged.

## Also

Drops the `category` fragment from product lines when it's the `"Product"` fallback — Shopify's
`productType` is empty on most of the live catalogue, so it padded the line without saying anything.
(#98 makes the same call for the `Product` schema; if both land, the check is worth unifying.)

## Verification

`lint` · `typecheck` · `build` pass. Build shows `/llms.txt` as **`○ (Static)` with `1h` revalidate**,
matching `sitemap.xml`.

| Check | Result |
| --- | --- |
| `GET /llms.txt` | 200, `Content-Type: text/plain; charset=utf-8` |
| Product lines | all 3 live products, prices via `formatPrice` (`R 300`, `R 210,99`, `R 100`) |
| `robots.txt` | AI group emitted with all 7 private paths repeated |
| Deploy-preview context | robots still `Disallow: /` with no AI group; llms.txt base URL follows `DEPLOY_PRIME_URL` |

## Two catalogue issues this surfaced (not code)

Both are visible in the generated file and will be visible to crawlers once live:

1. **"Test Product" (R 300, description "ABC") is an active, published product.** It appears in the
   sitemap, structured data and now llms.txt. Worth unpublishing before launch.
2. **"Glow Collagen Blend" has the wrong description** — it's a copy of the Plant Protein copy,
   describing 22g plant protein. Belongs with the content pass in #51.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -231,7 +231,7 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
 - **Structured data** (#14): schema.org payloads are built by pure functions in
   `src/lib/structured-data.ts` and rendered by `src/components/JsonLd.tsx` — `Product` +
   `BreadcrumbList` on `/store/[slug]`, `Organization` site-wide from `layout.tsx`.
-  **`JsonLd` escapes `<` to `<`, and that line is load-bearing** — Shopify descriptions are
+  **`JsonLd` escapes `<` to `\u003c`, and that line is load-bearing** — Shopify descriptions are
   merchant-controlled free text, so one containing `</script>` would otherwise close the tag early
   and have the rest parsed as markup.
   **The block is deliberately not nonced.** `proxy.ts` sets `script-src 'self' 'nonce-…'`, but a
@@ -245,6 +245,16 @@ For non-feature work substitute the type segment: `fix`, `chore`, `refactor`, `d
   for a related reason: Shopify's `vendor` is not fetched and reads "My Store" on most of the live
   catalogue. Absolute URLs come from `absoluteUrl()`, which passes Shopify CDN URLs through
   untouched.
+- **`llms.txt` and AI crawlers** (#14): `src/app/llms.txt/route.ts` serves the llmstxt.org
+  convention — an H1, a blockquote summary, then annotated links. **The filename is plural**;
+  #14 calls it "llm.txt" but `llms.txt` is the name crawlers actually look for. It is a Route
+  Handler rather than a file in `public/` so the product list tracks Shopify, and it reuses
+  `getProducts()` to inherit the hourly revalidate and the `products` tag — a product webhook
+  refreshes it too. Keep it `○ (Static)` in the build output; it reads no request data.
+  `robots.ts` names the AI crawlers explicitly and **repeats `PRIVATE_PATHS` for that group**.
+  That repetition is load-bearing, not copy-paste: per the robots.txt spec a crawler obeys only
+  the most specific group matching its name and ignores `*` entirely, so naming GPTBot without
+  the disallows would *widen* its access to `/account` rather than leave it unchanged.
 - **Page pattern**: a top-level route is a server `page.tsx` (`<Navbar /> … <Footer />` +
   `metadata`) that renders a `"use client"` `*Content.tsx` for interactivity (see
   `app/account`, `app/checkout`). Route protection lives in `src/proxy.ts` (Next 16 renamed the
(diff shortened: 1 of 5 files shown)


Oreginal:
## Verified locally and on the live deploy preview

Exercised against both catalogues, since they cover different branches of this code.

### Local, real Shopify catalogue (`NEXT_PUBLIC_USE_PLACEHOLDER_PRODUCTS=false`)

| Check | Result |
| --- | --- |
| `GET /llms.txt` | 200, `Content-Type: text/plain; charset=utf-8` |
| Products listed | all 3 live products, prices via `formatPrice` — `R 300`, `R 210,99`, `R 100` |
| Category omitted when it's the fallback | ✅ `Test Product` renders `R 300 · ABC`, no stray `· Product ·` |
| Category kept when real | ✅ `Hey Beautiful Plant Protein` keeps `· Supplements ·` |
| `robots.txt` | AI group emitted with all 7 private paths repeated |

### Live deploy preview

The preview runs the **development** catalogue (Netlify still has that flag on), which is useful
here — it exercises paths the 3-product live store cannot:

- **`Currently sold out.`** appears on `Iron + B12 Support`, exercising the `isSoldOut` branch.
  No live product is out of stock.
- Base URL correctly follows `DEPLOY_PRIME_URL`, not production's `URL`.
- **`robots.txt` on the preview is still just `User-Agent: * / Disallow: /`** — the AI group is
  correctly absent, since the whole preview is closed to crawlers.

`lint` · `typecheck` · `build` pass. Build output shows `/llms.txt` as **`○ (Static)` with `1h`
revalidate**, matching `sitemap.xml` — so it prerenders and the existing Shopify product webhook
refreshes it.

---

## Rebased onto #98, plus a small refactor

#98 merged while this was in flight and conflicted on `CLAUDE.md`. Rebased; both conventions entries
kept.

Two things came out of that, in a separate commit so they're easy to review independently:

1. **`meaningfulCategory()` extracted** to `src/lib/product.ts`. #98 and this PR had each
   independently discovered that `toProduct` substitutes the literal `"Product"` when Shopify's
   `productType` is empty, and each dropped it with its own copy of the comparison. One helper now,
   used by both `structured-data.ts` and `llms.txt/route.ts`.
2. **Fixed a doc typo that landed in #98.** The structured-data entry in `CLAUDE.md` read
   *"escapes `<` to `<`"* — a backslash was lost in transit, leaving a sentence that says nothing.
   It now reads `\u003c` as intended.

---

## The catalogue issues are worth a look before launch

Both were found against the **real** store, and both will be visible to crawlers in three places
each once the site is live — sitemap, JSON-LD, and now llms.txt:

- **`Test Product`** — R 300, description `ABC` — is an active, published product. It is already
  prerendered as a route and listed in the live sitemap.
- **`Glow Collagen Blend`** carries the Plant Protein description verbatim, down to "22g plant
  protein per serving".

Neither is a code problem. The first is a two-click unpublish in Shopify admin; the second belongs
with the content pass in #51.

